PR #18, opened 2016-04-11
[FLINK-3737] adding note about SOCKS proxy server

tlisonbee:
A test was failing because it needed a SOCKS proxy server when behind a firewall.  Adding note.

### DIFF
--- a/contribute-code.md
+++ b/contribute-code.md
@@ -187,7 +187,7 @@ The code is downloaded into a directory called `flink`.
 
 If you are behind a firewall you may need to provide Proxy settings to Maven and your IDE.
 
-Some tests communicate over IRC and need a [SOCKS proxy server](http://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) to pass.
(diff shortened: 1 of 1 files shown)


tlisonbee:
FYI, I also submitted a pull request to the source code with a note in the particular test that needed a SOCKS proxy, https://github.com/apache/flink/pull/1872

Thanks.

rmetzger:
Good change, +1 merge.
Maybe we should add the name of the test which needs the socks proxy. I don't think there are more tests connecting to some online service.

tlisonbee:
Sure, that makes sense.  I updated the language to use that test as an example (in case others are added in the future).

rmetzger:
Thx. Will merge it now.

rmetzger:
I've merged the change but forgot to include the magic string to close this pull request.
Can you close it manually?

tlisonbee:
Closing for rmetzger

rmetzger:
Thank you!

PR #40, opened 2016-10-14
Web cleanup and docker

greghogan:
I had sent @fhueske some small fixes back in April but had not gotten around to committing the changes. I had issues building the site locally so I replicated the Docker configuration from Flink docs, though with Ubuntu rather than CentOS as the latter uses Ruby 2.0.0 and was giving errors.


fhueske:
Thanks @greghogan.
I don't have experience with Docker, but the other changes look good.

PR #44, opened 2017-01-09
Flink site redesign

wints:
This PR includes substantial updates to the Flink site (as were covered in the dev@flink.apache.org thread _[DISCUSS] Proposed updates to Flink project site_). 

Here are a few of the most sweeping changes:
• Site layout: left-side nav instead of top nav
• Homepage redesign: New graphic, add 'Powered by' Carousel, fewer links to other pages than the current site
• Add 'Introduction to Flink' writeup, geared toward brand new users (this is where the 'Features' section now lives, too)

Note that no files from the old site were deleted, so external links to pages that aren't included in the new nav bar will still work as expected.


wints:
@rmetzger @uce @fhueske

StephanEwen:
Could you post a screenshot or a link to a preview here?

wints:
@StephanEwen Sure thing. Here's a preview: https://wints.github.io/flink-web/

rmetzger:
Thanks a lot, great work!

I think the change adds value for our users.
I'm wondering if there's a way to make it a bit more prominent that we are part of the Apache Software Foundation. Look for example at the [hadoop](hadoop.apache.org) page. They've linked the Foundation multiple times at the top of the page.
On our page, we don't have "Apache" in our logo, and the only link is in the footer.
The old page also wasn't very good at this

wints:
@rmetzger That's a good idea, but what are your thoughts on making that a separate project that we take on after the redesigned site is live? I volunteer to put together a few ideas to share (such as a modified logo or header bar similar to what's on the [Mesos](http://mesos.apache.org/)) site.

rmetzger:
+1 to fix that later on.

aljoscha:
I couldn't find a link to the wiki. Is it available but hidden somewhere in a sub-page?

wints:
@aljoscha Leaving out a top-level link was done intentionally--we do link specific pages in the wiki (e.g. the longer "Powered By" here https://wints.github.io/flink-web//poweredby.html and the FLIP section here https://wints.github.io/flink-web//how-to-contribute.html#propose-an-improvement-or-a-new-feature). 

It'd be easy to add this https://cwiki.apache.org/confluence/display/FLINK/Apache+Flink+Home to the left-hand nav as its own link or as a sub-selection under "Documentation" if we want this to still be visible. Let me know your thoughts.

aljoscha:
I think at least having a link to the top-level wiki on the community page would be good since there is also other stuff in the wiki that is interesting for some people. I don't think we need it in the side nav.

wints:
@aljoscha Makes sense--I just pushed a change to include the wiki link on the Community page. I am concerned that there's some potentially outdated info in the wiki (e.g. articles that haven't been updated since 2015), and so a wiki cleanup could also be a worthwhile project separate from this one.

vasia:
Hi @wints,
thanks a lot for your work on the website. It looks much nicer than before :)
Just a couple of comments from my side (if you have time to consider):
- I would add the "download flink" button next to the "introduction to flink" button. When browsing the new website, I first scrolled down until the end and then looked at the menu, and then found the download button. I suspect that many users who visit the page would like to find how to download flink immediately.
- I like the banner with logos on the bottom, but maybe we should add a header that says "powered by" or similar. I'm afraid the logos can lead visitors to believe that these are sponsors.

wints:
@vasia thank you for the feedback! Both are helpful suggestions. 

- Agreed that the 'Download Flink' button being too difficult to locate--for starters, I moved it to the top of the nav menu. I think this makes it a lot more prominent. 

- As for the 'Powered by Flink' section on the homepage: I added a header and linked the header itself to the full 'Powered by' page.

rmetzger:
I want to add https://github.com/kempa-liehr/DSVMs/issues/1 to the 3rd party / ecosystem page. I don't think it makes sense to add it to the old page, so I'm putting it here.
@wints please only add this if you need to do another change to the website. Otherwise, I can add the link to the ecosystem page, once your page rework has been merge.

PR #45, opened 2017-01-19
Added a page that redirects to the latest stable docs. 

alpinegizmo:
We can link to this page from older versions of the docs.


rmetzger:
I'll merge the change

rmetzger:
Can you close the PR?
Looks like it works: http://flink.apache.org/q/stable-docs.html

PR #47, opened 2017-02-15
get the poweredby page to render properly

alpinegizmo:
The poweredby page often looks something like this, which makes me sad:

screen shot 2017-02-15 at 6 13 51 pm

As best I can tell, there's no way to fix this with CSS, but a bit of JS does the trick. This finds the tallest of the company blurbs and sets all of their heights to that max height, allowing everything to float properly.

### DIFF
--- a/poweredby.md
+++ b/poweredby.md
@@ -12,6 +12,7 @@ title: "Powered by Flink"
    th, td {
    padding: 10px;
    }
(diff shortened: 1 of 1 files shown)


uce:
Good catch. Do you know why this does not happen for the other boxes? If possible, I would really love to have this fixed w/o any JS.

PR #50, opened 2017-03-14
Add some text on bugfix policy

rmetzger:
I added some text on the bugfix policy on the download page (the FAQ is horribly outdated)
updatepol

And a warning on how to use the mailing list:
warn


tzulitai:
LGTM! Merging ..

PR #51, opened 2017-03-28
Add new blog post on Table & SQL API

wints:
Just one change: adding a new blog post on the Table & SQL API authored by @twalthr.

### DIFF
--- a/_posts/2017-03-23-release-1.1.5.md
+++ b/_posts/2017-03-23-release-1.1.5.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title:  "Apache Flink 1.1.5 Released"
-date:   2017-03-22 18:00:00
+date:   2017-03-23 18:00:00
 categories: news
 ---
 
--- a/content/blog/feed.xml
+++ b/content/blog/feed.xml
@@ -7,6 +7,183 @@
 <atom:link href="http://flink.apache.org/blog/feed.xml" rel="self" type="application/rss+xml" />
 
 <item>
+<title>From Streams to Tables and Back Again: An Update on Flink&#39;s Table &amp; SQL API</title>
+<description>&lt;p&gt;Stream processing can deliver a lot of value. Many organizations have recognized the benefit of managing large volumes of data in real-time, reacting quickly to trends, and providing customers with live services at scale. Streaming applications with well-defined business logic can deliver a competitive advantage.&lt;/p&gt;
+
+&lt;p&gt;Flink’s &lt;a href=&quot;https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/datastream_api.html&quot;&gt;DataStream&lt;/a&gt; abstraction is a powerful API which lets you flexibly define both basic and complex streaming pipelines. Additionally, it offers low-level operations such as &lt;a href=&quot;https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/stream/asyncio.html&quot;&gt;Async IO&lt;/a&gt; and &lt;a href=&quot;https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/stream/process_function.html&quot;&gt;ProcessFunctions&lt;/a&gt;. However, many users do not need such a deep level of flexibility. They need an API which quickly solves 80% of their use cases where simple tasks can be defined using little code.&lt;/p&gt;
+
+&lt;p&gt;To deliver the power of stream processing to a broader set of users, the Apache Flink community is developing APIs that provide simpler abstractions and more concise syntax so that users can focus on their business logic instead of advanced streaming concepts. Along with other APIs (such as &lt;a href=&quot;https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/libs/cep.html&quot;&gt;CEP&lt;/a&gt; for complex event processing on streams), Flink offers a relational API that aims to unify stream and batch processing: the &lt;a href=&quot;https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/table_api.html&quot;&gt;Table &amp;amp; SQL API&lt;/a&gt;, often referred to as the Table API.&lt;/p&gt;
+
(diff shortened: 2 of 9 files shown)


wints:
@fhueske

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+This blog post summarizes the current status of Flink’s Table API and showcases some of the recently-added features in Flink 1.2.0. Among the features presented here are the unified access to batch and streaming data, data transformation, and window operators.
+The following paragraphs are not only supposed to give you a general overview of the Table API, but also to illustrate the potential of relational APIs in the future.
+
+Because the Table API is built on top of Flink’s core APIs, [DataStream](https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/datastream_api.html) or [DataSet](https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/batch/index.html) programs can be converted to a Table and vice-versa without much overhead. Hereafter, we show how to create tables from different sources and specify programs that can be executed locally or in a distributed setting. In this post, we will use the Scala version of the Table API, but there is also a Java version as well as a SQL API with an equivalent set of features.
```
fhueske: DataStream and DataSet programs can be converted into a Table -> DataStreams and DataSets can be converted into a Table

fhueske: "... but there is also a Java version as well ...", remove either "also" or "as well"?

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+The Table API comes with a large set of built-in functions that make it easy to specify  business logic using a language integrated query (LINQ) syntax. In the example above, we filter out customers with invalid names and only select those that updated their preferences recently. We convert names to lowercase for normalization. For debugging purposes, we convert the table into a DataStream and print it.
+
+The `CsvTableSource` supports both batch and stream environments. If the programmer wants to execute the program above in a batch application, all he or she has to do is to replace the environment via `ExecutionEnvironment` and change the output conversion from `DataStream` to `DataSet`. The Table API program itself doesn’t change.
+
+In the example, we converted the table program to a data stream of `Row` objects. However, we are not limited to row data types. The Table API supports all types from the underlying APIs such as Java and Scala Tuples, Case Classes, POJOs, or generic types that are serialized using Kryo. Let’s assume that we want to have regular object (POJO) with the following format instead of generic rows:
+
+```scala
+class Customer {
```
fhueske: use a case class instead of a POJO?

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+We can use the following table program to convert the CSV file into Customer objects. Flink takes care of creating objects and mapping fields for us.
+
+```scala
+val ds = tEnv
+  .scan("customers")
+  .select('id, 'name, 'last_update as 'update, parseProperties('prefs) as 'prefs)
+  .toDataStream[Customer]
+```
+
+Even though Flink’s Table API is shipped with a large set of built-in functions, it is often necessary to define custom user-defined scalar functions in this example to parse the preferences into a Java Properties object. The following code snippet shows how easily we can implement a scalar function.
```
fhueske: I would make it more clear that the query above uses a UDF. Right now, the connection between the two parts seems to be missing.

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+
+Let’s assume that our customer data from the example above is an event stream of updates generated whenever the customer updated his or her preferences. We assume that events come from a TableSource that has assigned timestamps and watermarks. The definition of a window happens again in a LINQ-style fashion. The following example could be used to count the updates to the preferences during one day.
+
+```scala
+table
+  .window(Tumble over 1.day on 'rowtime as 'w)
+  .groupBy('id, 'w)
+  .select('id, 'w.start as 'from, 'w.end as 'to, 'prefs.count as 'updates)
+```
+
```
fhueske: The point that timestamps & watermarks have to be assigned is mentioned several times.

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+```scala
+// using processing-time
+table.window(Tumble over 100.rows as 'manyRowWindow)
+// using event-time
+table.window(Session withGap 15.minutes on 'rowtime as 'sessionWindow)
+table.window(Slide over 1.day every 1.hour on 'rowtime as 'dailyWindow)
+```
+
+Since batch is just a special case of streaming (where a batch happens to have a defined start and end point), it is also possible to apply all of these windows in a batch execution environment. Without any modification of the table program itself, we can run the code on a DataSet given that we specified a column named “rowtime”. This is particularly interesting if we want to compute exact results from time-to-time, so that late events that are heavily out-of-order can be included in the computation.
```
fhueske: some of these windows on batch are not included in Flink 1.2

_posts/2017-03-28-table-sql-api-update.md
twalthr: "is often necessary to define custom user-defined scalar functions. In the above example we use a user-defined function `parseProperties`..."

_posts/2017-03-28-table-sql-api-update.md
```diff
@@ -0,0 +1,195 @@
...
+title:  "From Streams to Tables and Back Again: An Update on Flink's Table & SQL API"
+excerpt: "<p>Broadening the user base and unifying batch & streaming with relational APIs</p>"
+date:   2017-03-27 12:00:00
+author: "Timo Walther"
+author-twitter: "twalthr"
+categories: news
+---
+Here’s the thing about stream processing: it offers a lot of value, but it sure can be hard. Many organizations have recognized the benefit of managing large volumes of data in real-time, reacting quickly to trends, and providing customers with live services at scale. Streaming applications with well-defined business logic can deliver a competitive advantage.
+
+But streaming concepts such as windows, event time, out-of-order event handling, and state management are challenging for many users who could benefit from streaming. Most streaming APIs require deep Java and Scala knowledge, especially when it comes to efficient programming, thus limiting the range of users who can interact with modern streaming frameworks.
```
wints: @twalthr I think that replacing the second paragraph with your rewrite would work well

_posts/2017-03-28-table-sql-api-update.md
fhueske: Was thinking about starting the paragraph with something like this: 

> "You might have noticed that the query above uses a function to parse the properties field. Even though Flink's Table API ..."

wints: Added this @fhueske

wints:
@twalthr I just pushed an updated version based on your and @fhueske's comments here.

PR #52, opened 2017-03-29
Add blog post 'Continuous Queries on Dynamic Tables'

fhueske:
This is a blog post I wrote together with @shaoxuan-wang and Xiaowei Jiang about Dynamic Tables.

We proposed dynamic tables in this [design doc](https://docs.google.com/document/d/1qVVt_16kdaZQ8RTfA_f4konQPW4tnl8THw6rzGUdaqU) and discussed the feature on the mailing list.


rtudoran:
@fhueske 

Thanks for sharing this blog post.

What I am lacking the understanding is the implications of not using the dynamic tables. We have discussed this on the mailing list that there are scenarios in which you might just want to use the Stream SQL to generate the stream topology. For such simple cases - is there an implication of bring dynamic tables? I would expect that such scenarios are still possible.

Another question that can be clarified is if dynamic tables will be similar with stream source/sinks? My understanding on the concept is that it would be the case. I think it could be made a bit more explicit.

Regarding the discussion that mentions that once you create a dynamic table that is updated by a continuous query, you can use SQL to query the dynamic table. I think this can be clarified a bit more. Assuming you define a query to count the elements over the last hour. You have a dynamic table in which you put the results of this query (append table?). Then to get the current count you would have to execute another query to look into the table?

Then if you would like (or maybe this is for the follow-up post) a discussion about the queries that make sense or do not make sense can be done in more details. This is only briefly mentioned. I have a couple of such issues that I discuss also in the JIRA issues that I opened for the other SQL statements. Other can be define as well.

fhueske:
Hi @rtudoran,

the concept of Dynamic Tables is a super set of what the Table API allows to do right now.
At the moment, you can consider all streaming tables to be created in *Append Mode* and all implemented operators only apply append operations on the result tables (projection, filter, windows without late data handling). So right now, the Table API implements a subset of what the blog post describes.

I hope that answers your questions about the relationship between dynamic tables and the current state.

Thanks, Fabian

rtudoran:
@fhueske thanks for the explanation - it is definitely helpful - however still translate this into my mind into something like a sink where all resulting logic from the SQL operations append data.
Is this correct at least to some extend?

fhueske:
In the generic case, a dynamic table can be modified in any way insert row, update row, delete row. However, there are special cases (like the set of currently supported operations) that only result in append operations. In this case, the resulting stream can be a stream of the records which are inserted into the logical table.

Tables which are modified by insert/update/delete are converted into changelog streams as described in the post.

It is important to understand that dynamic tables are just a logical concept, the queries will be translated into DataStream programs just like before. However, we need to keep more state for certain operations. For instance a grouped aggregation (without window) will need to keep the full result in state to be able to update the results when more data arrives.

rtudoran:
@fhueske - now i got it! Thanks a lot!
 IMHO you should add also this explanation somewhere in the blogpost.

fhueske:
Thanks for the feedback. I'll try to make that more clear.

rmetzger:
I think this PR can be closed?

fhueske:
Yes

PR #53, opened 2017-03-30
Update footer for 2017

wints:
Super small change. Updating the footer on the site so it shows "2014-2017".


wints:
@rmetzger @uce

rmetzger:
+1

PR #57, opened 2017-05-05
fix mis-spelling real-word to real-world.

sunjincheng121:
If i understand correctly the `real-word` in the sentence "Many real-word data sets that are traditionally thought of as bounded or “batch” data are in reality unbounded datasets" has a mis-spelling. we should change `real-word` to `real-world`.


wints:
+1. good catch and thanks for fixing this, @sunjincheng121!

rmetzger:
will merge

PR #71, opened 2017-06-30
Update Gelly documentation link

njzhuyuqi:
Update Gelly documentation link on the flink news web page address http://flink.apache.org/news/2015/08/24/introducing-flink-gelly.html, change to the newest documentation of Gelly.

### DIFF
--- a/_posts/2015-08-24-introducing-flink-gelly.md
+++ b/_posts/2015-08-24-introducing-flink-gelly.md
@@ -322,7 +322,7 @@ certain song.
 Graph<String, NullValue, Integer> userSongGraph = Graph.fromTupleDataSet(validTriplets, env);
 ```
 
-Consult the [Gelly guide](https://ci.apache.org/projects/flink/flink-docs-master/libs/gelly_guide.html) for guidelines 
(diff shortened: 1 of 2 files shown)


zhangminglei:
LGTM, Thanks for @njzhuyuqi contribution.

tzulitai:
Hi @njzhuyuqi, thanks for the PR.
I think we actually need to update the .md file that generates this html, not change the generated html directly. Otherwise every time the site is rebuilt, the link will still be overwritten with the incorrect link.

njzhuyuqi:
hi,  tzulitai , thanks for pointing out this,  I have updated the PR.

zhangminglei:
@njzhuyuqi I believe that you just change the .md file is enough, no need to change html file instead.

tzulitai:
+1, merging!

@zhangminglei it's fine, as long as the markdown file is modified. If the html file isn't modified also in the PR, it would just be a matter of rebuilding the site once the PR is merged.

njzhuyuqi:
yes, i see, thanks zhangminglei and tzulitai.

PR #75, opened 2017-07-29
Add Flink 1.3.2 release

aljoscha:
This is to be finalised with the correct release data after the release.


tzulitai:
Should we include text to notify about the following user affecting changes?
- https://issues.apache.org/jira/browse/FLINK-7174
- https://issues.apache.org/jira/browse/FLINK-6365

tzulitai:
Maybe we should also state that it is not recommended to migrate from savepoints taken with 1.3.0 / 1.3.1 if the Kafka consumer was used as the source, since due to https://issues.apache.org/jira/browse/FLINK-7143 the assignment may be incorrect and have gone unnoticed.

aljoscha:
Yes, do you have sth in mind for the first two issues? I was just about to write something for the Kafka-inconsistent-state thing.

tzulitai:
@aljoscha I've left "Release Notes" in the tickets of FLINK-7174 and FLINK-6365 which I think can be directly used as text for the announcement.

aljoscha:
cool, thanks!

aljoscha:
@tzulitai I updated the release notes and also updated the notice about the Kafka bug.

tzulitai:
@aljoscha thanks!
I like the additions a lot :)
+1, LGTM (after the release, of course :) )

PR #76, opened 2017-08-03
[FLINK-7361] upgrade to Jekyll 3 to fix build with Ruby 2.4

NicoK:
This PR updates all rubygem dependencies to newer versions so that the suite builds again when using ruby 2.4.

Note that a few anchors and HTML characters changed but the overall site should be the same with the (included) minor fix in the navbar. Please double-check though!


zentol:
I can build the site with ruby 2.4 as is, without requiring any changes to the dependencies.

Could you paste the output of `bundle install --full-index --path .rubydeps` on the current `asf-site` branch?

NicoK:
seems, I cannot reproduce it anymore since a new `yajl-ruby` (version 1.2.2) is pulled in now, at least when starting from a clean `.rubydeps`.

Should we leave this PR then as a "feature" request of bumping dependencies to be more future-proof there?

zentol:
The conclusion of FLINK-6302 was to only increase dependencies if it is necessary.

NicoK:
ok, I'll close this one and create a new PR for the "lighter" change fixing ruby 2.4 builds...
if at any point in time we need Jekyll 3, we can re-use this one (if we remember)

PR #77, opened 2017-08-03
[FLINK-7363] add hashes and signatures to the download page

NicoK:
As part of the releases, we also generate MD5 hashes and cryptographic signatures but neither link to those nor do we explain which keys are valid release-signing keys.

This PR adds them along with a small introduction and verification instructions.

Please note that this PR is based on #76 because I could not build the website without it.


zentol:
Changes look good, will merge it after a rebase&rebuild.

PR #79, opened 2017-08-08
[FLINK-7390] Fix 1.2.0-1.3.2 source download links

zentol:


content/downloads.html
```diff
@@ -258,11 +258,11 @@ <h2 id="all-releases">All releases</h2>
...
-  <li>Flink 1.2.0 - 2017-02-06 (<a href="http://www.apache.org/dyn/closer.lua/flink/flink-1.2.0/flink-1.2.0-src.tgz">Source</a>, <a href="http://archive.apache.org/dist/flink/flink-1.2.0/">Binaries</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.2/">Docs</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.2/api/java">Javadocs</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.2/api/scala/index.html">ScalaDocs</a>)</li>
+  <li>Flink 1.3.2 - 2017-08-05 (<a href="http://archive.apache.org/dist//flink/flink-1.3.2/flink-1.3.2-src.tgz">Source</a>, <a href="http://archive.apache.org/dist/flink/flink-1.3.2/">Binaries</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.3/">Docs</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.3/api/java">Javadocs</a>, <a href="http://ci.apache.org/projects/flink/flink-docs-release-1.3/api/scala/index.html">ScalaDocs</a>)</li>
```
NicoK: - there is a double slash now, after `dist` which we could avoid
- also, if we're changing the URLs, why not go with https now?

zentol: will fix it

zentol:
For archive.apache.org we use direct links to `archive.apache.org` because the mirror listing at `apache.org/dyn/closer.cgi` is unreliable as hell for archived releases. Some mirrors contain them, while others don't. (i.e, some only mirror dist.apache.org, and not archive.apache.org) (And there doesn't appear to be any guarantee that they are available at _any_ mirror)

As for the latest release, the apache guidelines say to refer to the mirror sites for latest releases. I would love to have a link that automatically picks the closest mirror, but I couldn't find a way to get it.

zentol:
@NicoK Could you take another look?

PR #81, opened 2017-08-28
[FLINK-6860] Update apache beam info(remove incubating)

yew1eb:
To remove the word incubating and update the link, --Apache Beam has graduated as a top-level project.


zentol:
merging.

PR #87, opened 2017-10-04
Update community.md

danibentrup:


tzulitai:
LGTM, merging this! Thanks @danibentrup.

tzulitai:
Thanks for extending this list @danibentrup!

The changes LGTM, with one minor concern with the addition of the "data Artisans Product Announcement" talk. IMO, we should avoid adding commercial product announcements of this kind to this list.

Will merge this while removing that particular from the list. Please let me know if you disagree with this action.

zentol:
@danibentrup Could you close the PR? (it has been merged already)

PR #94, opened 2017-11-22
Add post to announce 2017 Apache Flink user survey

wints:
The 2017 Apache Flink user survey is open for responses, and this short post announces the survey on the Flink project site. @aljoscha


wints:
The survey will close down at the end of the day today Mon, Nov 27, and so this announcement post won't be necessary. Closing the PR.

PR #98, opened 2017-12-23
Update introduction to ES sink

yew1eb:


zentol:
@yew1eb I merged the PR, could you close it?

PR #105, opened 2018-04-23
[FLINK-9222] add gradle quickstart script

NicoK:
This adds a quickstart script to setup a project using `gradle` for management. It handles shading/shadow jars correctly, works for both snapshot and release versions, allows application and test runs with embedded Flink, and allows the project to be used by IDEs (tested with IntelliJ). Part 2 of this PR will add and refer to it in the Flink docs.

### DIFF
--- a/q/gradle-quickstart.sh
+++ b/q/gradle-quickstart.sh
@@ -1,0 +1,246 @@
+#!/bin/bash
+
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+declare -r TRUE=0
+declare -r FALSE=1
+
+# takes a string and returns true if it seems to represent "yes"
+function isYes() {
+  local x=$1
+  [ $x = "y" -o $x = "Y" -o $x = "yes" ] && echo $TRUE; return
+  echo $FALSE
+}
+
+function mkDir() {
+  local x=$1
+  echo ${x// /-} | tr '[:upper:]' '[:lower:]'
+}
+
+function mkPackage() {
+  local x=$1
+  echo ${x//./\/}
+}
+
+defaultProjectName="quickstart"
+defaultOrganization="org.myorg.quickstart"
+defaultVersion="0.1-SNAPSHOT"
+defaultScalaBinaryVersion="2.11"
+defaultFlinkVersion="1.6-SNAPSHOT"
+
+echo "This script creates a Flink project using Java and Gradle."
+
+while [ $TRUE ]; do
+
+  echo ""
+  read -p "Project name ($defaultProjectName): " projectName
+  projectName=${projectName:-$defaultProjectName}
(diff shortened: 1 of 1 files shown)


q/gradle-quickstart.sh
```diff
@@ -0,0 +1,222 @@
...
+
+
+sourceCompatibility = javaVersion
+targetCompatibility = javaVersion
+tasks.withType(JavaCompile) {
+	options.encoding = 'UTF-8'
+}
+
+applicationDefaultJvmArgs = ["-Dlog4j.configuration=log4j.properties"]
+
```
knaufk: Add 
```
task wrapper(type: Wrapper) {
    gradleVersion = <some_version> 
}
```
This way if the user sets up and uses the gradle wrapper (as is best practices), the correct version (=the version we developed against) is used as opposed to the version the user has currently installed on his system.

q/gradle-quickstart.sh
```diff
@@ -0,0 +1,222 @@
...
+// declare where to find the dependencies of your project
+repositories {
+    mavenCentral()
+    maven { url "https://repository.apache.org/content/repositories/snapshots/" }
+}
+
+// NOTE: We cannot use "compileOnly" or "shadow" configurations since then we could not run code
+// in the IDE or with "gradle run". We also cannot exclude transitive dependencies from the
```
knaufk: In IntellijC2018.1 it works with compileOnly dependencies and with 
```
run.classpath sourceSets.main.compileClasspath
```
we can change the classpath for the `run`task.

knaufk: This should also solve the problem with the transitive dependencies, right? We set them to compileOnly and they are not included in the shadowJar including their transitive dependencies.

NicoK: For dependencies in `compileOnly` transitive dependencies are not included, but this is about the (unwanted) transitive dependencies of jars that should end up in the shadowJar and this part is not yet possible otherwise.

q/gradle-quickstart.sh
```diff
@@ -0,0 +1,226 @@
...
+
+mkdir -p ${directoryName}
+cd ${directoryName}
+
+# Create the README file
+
+cat > README <<EOF
+A Flink application project using Java and Gradle.
+
+To package your job for submission to Flink, use: 'gradle shadowJar'
```
twalthr: Can you add the location of the built jar here as well?

q/gradle-quickstart.sh
```diff
@@ -0,0 +1,226 @@
...
+    flinkShadowJar.exclude group: 'org.slf4j'
+    flinkShadowJar.exclude group: 'log4j'
+}
+
+// declare the dependencies for your production and test code
+dependencies {
+    compile "org.apache.flink:flink-java:\${flinkVersion}"
+    compile "org.apache.flink:flink-streaming-java_\${scalaBinaryVersion}:\${flinkVersion}"
+
+    // Add connector dependencies here.
```
twalthr: Not only `connector`. In general I would make this section more prominent. Maybe using
```
// --------------------------------------------------------------
// Add dependencies here that should NOT be part of the 
// shadow jar and are provided in the lib folder of Flink
// --------------------------------------------------------------
...
// --------------------------------------------------------------
// Add dependencies here that should be part of the shadow jar
// --------------------------------------------------------------
```

q/gradle-quickstart.sh
NicoK: I highly doubt that a gradle user will not know where to find the file, but why not...

NicoK:
Actually, I'd keep the difference in the case of the project name and the directory: we do print where the project is created and therefore it is fine (there are some more differences between project name and directory name because of file system limitations anyway)

I fixed the other issues (in both PRs)

NicoK:
FYI: I just also verified that `./gradlew test` works with the given project file, after adding a unit test and enabling junit4

StephanEwen:
What do you think about adding this to the main Flink repository, like the Java/Scala project templates?

NicoK:
@StephanEwen are you suggesting to put this into a separate sub-module of `flink-quickstart` in apache/flink? We can't publish these to maven central in the same way though (similar to the `sbt` quickstart) and this script would have to download all files from git(?), interpret it as a template and replace placeholders?

Not quite sure whether this is feasible:

PR #121, opened 2018-08-28
 [FLINK-10238] Link nightly snapshot builds from the Download page

StephanEwen:
How to access the nightly builds is described in a bit of a hidden place, but is a question frequently popping up.

This adds the nightly builds to the download page, with a prominent disclaimer that they are not endorsed releases.

This also removes broken references to no longer existing builds bundling Hadoop v1.x.


zentol:
+1

StephanEwen:
Thanks for checking this out.
Merging...

PR #131, opened 2018-10-30
[FLINK-2028] Remove snapshot link

zentol:
This PR removes the snapshot build from the download page according to [ASF policy](http://www.apache.org/legal/release-policy.html#what).

> Do not include any links on the project website that might encourage non-developers to download and use nightly builds, snapshots, release candidates, or any other similar package.


StephanEwen:
I would like to revert this commit, and then have a proper discussion whether to remove this.

I am very aware of the cited rules and that's why there is a big disclaimer above the links.

We need to involve the people that created the original contents before removing it. Everything else does not work in the long run, we just keep adding and removing each others' contents over different ideas of what should be there.

uce:
Sebb's comments in the [ticket](https://issues.apache.org/jira/browse/FLINK-2028) indicate that it should be OK to have it on a page that is only advertised to Flink developers. @zentol had asked whether the Wiki would be acceptable, but forgot to follow up.

According to Sebb the Downloads page seems to be not acceptable though. Personally, I'm happy to have it on the Downloads page, but I would suggest to first ping the ASF or Sebb before reverting the commit.

tillrohrmann:
I know too many cooks spoil the broth, but I think the contribution guide would not be a good place. Maybe the "Community & Project Info" page would be better suited. Or one adds a dedicated page for Flink developers where we also link the committer guidelines and other developers related resources.

uce:
I think a separate page for Flink developers is a great idea. It would also make the case for the snapshot pretty obvious.

StephanEwen:
The experience from last time is that when they are too hidden, no one will access them and we will get repeated questions about where to find them.

The nightly builds are also interesting to users that want to check whether an upcoming feature is interesting/relevant to them, or to give early feedback (which is again valuable to the community).

I would like to check with sebb if an explicit disclaimer would actually work.

PR #141, opened 2018-12-18
Add 1.6.3 release

tzulitai:
Adds release announcement and download link for 1.6.3.


tzulitai:
Thanks for the review @tillrohrmann.
I'll fix `_config.yml` while merging.

PR #144, opened 2019-01-02
[FLINK-11252] Add scala 2.12 download column 

zentol:
This PR modifies the download page to feature Scala 2.11 / Scala 2.12 columns with their respective download links, instead of listing each scala-variant of a release as a separate item.

Scala variants are now defined in `_config.yaml` as fields of a particular release (effectively acting as Map entries) and queried in `downloads.md`.

new_downloads

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -50,7 +50,7 @@ source_releases:
 
 stable_releases:
   -
-      name: "Apache 1.7.1 Flink only"
+      name: "Apache Flink 1.7.1 only"
       scala_211:
           id: "170-download-hadoopfree_211"
           url: "https://www.apache.org/dyn/closer.lua/flink/flink-1.7.1/flink-1.7.1-bin-scala_2.11.tgz"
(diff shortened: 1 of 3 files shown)


aljoscha:
Looks good. We should merge this ASAP because I think there are already users that are a bit confused by this.

aljoscha:
Side note: in "Apache Flink xxx only" the only looks a bit confusing. I only understood what it meant once I read the "with Hadoop" parts.

zentol:
Yes I dislike the `only` bit as well, but I don't have a better suggestion :(

zentol:
except maybe `without hadoop`

zentol:
I'll merge it as is since the `only` classifier has been there for a while.

aljoscha:
👌 

We can simply remove the "only" then its just "Download Flink 1.7.1" and "Download Flink 1.7.1 with Hadoop 2.8.0" or "bundled with Hadoop 2.8.1"

PR #147, opened 2019-01-08
 [FLINK-11269] Support listing optional components 

zentol:
This PR adds support for listing optional components. We can use this to provide convenient access to metric reporters, sql formats/connector and shaded-hadoop jars.

Previews:

**Collapsed**:
optional_collapsed
**Expanded**:
optional_expanded

As seen components are grouped by their category in a collapsible block. Each component has it's own table, listing all artifacts for each Flink version. Components may be scala dependent.

Components are defined in _config.yml. This PR adds the SQL formats to the list of optional components as an example of non-scala dependent artifacts. The _config.yml also contains an commented-out example for a scala-dependent component.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -21,6 +21,61 @@ FLINK_GITHUB_URL: https://github.com/apache/flink
 FLINK_CONTRIBUTORS_URL: https://cwiki.apache.org/confluence/display/FLINK/List+of+contributors
 FLINK_GITHUB_REPO_NAME: flink
 
+optional_components:
+  -
+    name: "Avro"
+    category: "SQL Formats"
+    scala_dependent: false
+    versions:
+      -
+        version: "1.7.1"
+        id: 171-sql-format-avro
+        url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-avro/1.7.1/flink-avro-1.7.1.jar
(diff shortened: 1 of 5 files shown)


zentol:
So I put this on the main download page to keep things in a central place, and make it easier to discover for users, particularly for when we drop hadoop-specific releases.

We could however also simply link to the hadoop/metrics/sql documentation and let them manage the actual download links, like what SQL does right now [here](https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/table/connect.html#dependencies).
This would trivially address the issue regarding older releases.

I'm wondering though whether the documentation really is the correct place for download links.

tillrohrmann:
My gut feeling tells me that the documentation is not the right place for this. I think it would be a good idea to deduplicate information and either collect it on the downloads page or an optional dependency page which is linked from the downloads page.

zentol:
ok cool, so we're agreeing on that point.

I suggest the following: I'll find a more intuitive design for collapse/expand and merge the PR as is. With the current # of items it still fits nicely on the download page; we can revisit this once we have a grasp on how much we want to offer here.
As for old versions, again I would defer this to a later point. I haven't thought about it, but for the foreseeable future we can keep old versions around without overloading the page.

zentol:
There's now a plus sign for each category:
optional_intuitive

PR #149, opened 2019-01-24
Add readme, notice, license and update the review guide for the bot

rmetzger:


NOTICE
```diff
@@ -0,0 +1,6 @@
+Apache Flink
+Copyright 2014-2018 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
```
zentol: I wouldn't consider the documentation a distribution and hence we don't need a NOTICE file.

If we do add it then we should follow a similar pattern as we do in the main repo and list all shipped dependencies, covering images and javascript bits.

reviewing-prs.md
```diff
@@ -91,9 +91,11 @@ This is the detailed code review of the actual changes, covering:
 
 Some code style guidelines can be found in the [Flink Code Style Page]({{ site.baseurl }}/contribute-code.html#code-style)
 
-## Review Template
+## Review with the @flinkbot
 
-The "Review Progress" section is posted at the end of the description of every new pull request.
+The Flink community is using a service called [@flinkbot](https://github.com/flinkbot) to help with the review of the pull requests.
```
zentol: I guess this is a "chicken or the egg" problem, but the bot isn't active yet is it?

reviewing-prs.md
```diff
@@ -107,4 +109,12 @@ The "Review Progress" section is posted at the end of the description of every n
...
 
-Reviewers should tick off the boxes (in order) to indicate the progress of the review.
\ No newline at end of file
+Reviewers can instruct the bot to tick off the boxes (in order) to indicate the progress of the review.
+
+For approving the description of the contribution, mention the bot with `@flinkbot approve contribution`. This works similarly with `consensus`, `architecture` and `quality`.
+
+For approving all aspects, put a new comment with `@flinkbot approve all` into the pull request.
+
+The syntax for requiring attention is `@flinkbot attention @username1 [@username2 ..]`
```
zentol: missing period at the end.

reviewing-prs.md
```diff
@@ -91,9 +91,11 @@ This is the detailed code review of the actual changes, covering:
 
 Some code style guidelines can be found in the [Flink Code Style Page]({{ site.baseurl }}/contribute-code.html#code-style)
 
-## Review Template
+## Review with the @flinkbot
 
-The "Review Progress" section is posted at the end of the description of every new pull request.
+The Flink community is using a service called [@flinkbot](https://github.com/flinkbot) to help with the review of the pull requests.
+
+The bot automatically posts a comment tracking the review progress for each new pull request:
```
zentol: I assume the bot will also cover flink-web and flink-shaded?

zentol:
Note that before activating the bot and merging any documentation regarding it I believe the Flink developers (or PMC) should be provided information on how the bot can be updated and/or disabled should the need arise.

reviewing-prs.md
rmetzger: No :( The current implementation only supports one repository.
I would propose to activate the bot on the apache/flink repo to fine-tune and debug it. Once we are happy with it, I can extend it to support multiple repositories.

rmetzger:
Good point. I will put the information into the wiki.
The only problem I see are the credentials for the virtual machine running it, and the GitHub account it is using. Maybe I reach out to Infra to see if there's a solution for that

zentol:
I believe we can setup a PMC-only wiki page to share the github credentials. It's a bit cumbersome though since we have to give each person permissions individually.

I suppose the VM is currently sponsored by dA?
Would be good to ask infra whether we can move this to some ASF hosted machine; these docs may be useful https://reference.apache.org/pmc/vms .
Alternatively, we could think about running the bot as a short-lived periodic job on jenkins.

PR #150, opened 2019-01-25
fix the images didn't display on github markdown pages.

itdamo:
fix the images didn't display on github markdown pages.  
`"{{ site.baseurl }}/img/` should be  replaced to `"/img/`. 

before: the image missing.
c8135806-ffa5-446c-97bb-dbdd4ec7ad48

after: the images was shown
f24a1099-153c-47d5-a05a-a31d2d7a62e9

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -9,7 +9,7 @@
           </button>
           <div class="navbar-logo">
             <a href="{{ site.baseurl }}/">
-              <img alt="Apache Flink" src="{{ site.baseurl }}/img/flink-header-logo.svg" width="147px" height="73px">
+              <img alt="Apache Flink" src="/img/flink-header-logo.svg" width="147px" height="73px">
             </a>
           </div>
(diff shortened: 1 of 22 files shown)


rmetzger:
@fhueske do you remember why we are using the $baseurl variable everywhere?
I assume the benefit of this approach is that we can host the page on other urls, where the page does not run on "root" (`/`).
I'm not sure if there is any practical use-case for this.

On the other hand, the entire website has been build using the $baseurl. I'm not sure how important it is to be able to read the website from GitHub's markdown preview to give up the flexibility.
I'm leaning towards rejecting this pull request because I don't see the benefit of it.

zentol:
That the page doesn't work properly on GitHub seems irrelevant to me, that's not what it's built for. There are plenty of other issues we'd have to solve if we'd wanted to fix that.

I'm in favor of rejecting the PR.

zentol:
Note that #153 actually seems to make use of `baseurl` to support multiple languages.

PR #151, opened 2019-01-28
Add user-zh@flink.apache.org mailing list into community section

rmetzger:
I thought about adding a warning box about english being the primary language of the community.
But I decided against this step and only add it if it is necessary.


rmetzger:
We should wait with the merging till the mailing list has been created :)

rmetzger:
ML has been created .. merging.

PR #152, opened 2019-01-30
[FLINK-11462] Fix 'Powered By' page on Firefox

patricklucas:


patricklucas:
cc @fhueske

fhueske:
Thanks for the PR @patricklucas!
I tried the fix on my machine, but unfortunately, it still seems to be broken on my Firefox.
All tiles have the same height, but do still overlap (see screenshot). 

2019-01-31-085952_1355x1217_scrot

Chromium works well.

rmetzger:
I have to confirm @fhueske's result: On the first page load, it works, but after a refresh, it looks like this:

image

patricklucas:
Haha, it's a bit crazy that refreshing breaks it, since it's just inline javascript that wouldn't change between refreshes. However I suspect that it's due to some _other_ javascript running before or after this that decides whether it's broken.

Using JS in this way is a giant hack anyway, so I'm going to see if I can just do this the right way.

fhueske:
Great, now it works on my machine!
I was look whether we need to add the new MIT-licensed dependency to some kind of license file, but didn't find anything. 
@rmetzger, do you know how dependencies are handled here?

rmetzger:
I don't think the license is an issue here, because we are not redistributing the code.
And MIT license is in general very friendly.
The minified file contains a link to the repo and mentions the license.
+1 to merge from a license point of view.

patricklucas:
Please hold, I just noticed a typo.

PR #153, opened 2019-02-05
[FLINK-11527] Support multiple languages for the framework of flink-web

wuchong:
## What is the purpose of the change

This changes update the framework to support English and Chinese at the same time.

I excluded the `/content` file changes to make the PR easy to review.

## Brief change log
* integrate the `jekyll-multiple-languages` plugin.
* upgrade jekyll to '3.0.5' in Gemfile, because `jekyll-multiple-languages` depends on it.
* add `_data/i18n.yml` and update `_include/navbar.html` to support Chinese & English for sidebar.
* copy all the `xxx.md` to `xxx.zh.md` but keep the original English contents.
* I didn't copy the `xxx.zh.md` for the `/blog` and `/_post` files. I'm not sure whether we need to translate them, so the link of "Flink Blog" in Chinese website still link to "http://0.0.0.0:4000/blog/"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,9 +21,11 @@ source 'https://rubygems.org'
 ruby '>=1.9.0'
 
 # Dependencies required to build the Flink docs
-gem 'jekyll', '2.5.3'
+gem 'jekyll', '3.0.5'
 gem 'kramdown', '1.10.0'
 gem 'pygments.rb', '0.6.3'
 gem 'therubyracer', '0.12.2'
 # explicitly require yajl-ruby (dependency of jekyll) in a version that works with Ruby 2.4
 gem 'yajl-ruby', '1.2.2'
+gem 'jekyll-multiple-languages', '2.0.3'
+gem 'jekyll-paginate', '1.1.0'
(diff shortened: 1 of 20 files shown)


fhueske:
Thanks for the PR @wuchong.
I'll have a look soon.

Cheers, Fabian

fhueske:
Hi @wuchong, the PR looks good.
One thought, do you think it would be helpful if the language switch would link to the same page (of the other language) instead of the start page. For example, if you are on the English community page, the language link would lead to the Chinese community page. 

It might not be super important for the website (because it is fairly small) but for the docs with many pages it might be helpful.

Thanks, Fabian

wuchong:
@fhueske , I think it's a good idea. I have updated the PR. Note that we don't translate the blog pages, so when you are on the English blog page, the language link would link to the Chinese home page.

fhueske:
Hi Jark, thanks for the update.
I noticed that on the Chinese pages page-internal links stay in the same language, while links to another page always point to the English version. I guess that's just how it is until the Chinese page gets translated and the links are adjusted.

+1 to merge the PR. 
Do you want to merge it or should I do it?

wuchong:
Hi @fhueske , you are right. The links will be adjusted when translating.

I will try to merge this.  :)

fhueske:
Thank you!

PR #174, opened 2019-02-24
Add blog post 'Monitoring Flink Applications 101'

knaufk:


fhueske:
Thanks for the great post @knaufk.
I noticed that the links to the documentation point to the "stable version" and hence might be broken if we change the documentation in future versions.
I'd change them to the docs of the 1.7 release which should be more stable.

I'll fix that and will merge the blog post later today.

Best, Fabian

fhueske:
Published the post.
Thanks @knaufk!

PR #180, opened 2019-03-01
Update Downloads page for Flink 1.8

aljoscha:
This PR does two things:

1. Refactor/reshuffle the Downloads page
2. Add download links for Flink 1.8

The first one is a bit of a bigger change. What I try to achieve with the change is to make everything there is for one release available in one place and easily discoverable. The previous version had a header `Latest stable release (v1.7.2)`, which contains all past binary releases, not just 1.7.2, where you have to hunt for your version in a Table. If you wanted to get the additional SQL components you had to be aware that they existed, scroll past the source release section, expand the optional components section and then hunt for the component that corresponds to your release.

With this change, the optional components are right with the binary release and clearly visible, same for the optional pre-bundled Hadoop.

For comparison, you can look at the current downloads page: https://flink.apache.org/downloads.html and these screenshots of the change:
screenshot 2019-03-01 at 10 44 02
screenshot 2019-03-01 at 10 44 11
screenshot 2019-03-01 at 10 44 16

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,7 +10,6 @@ url: https://flink.apache.org
 DOCS_BASE_URL: https://ci.apache.org/projects/flink/
 
 FLINK_VERSION_STABLE: 1.7.2
-FLINK_VERSION_HADOOP_2_STABLE: 1.7.2
 FLINK_VERSION_STABLE_SHORT: 1.7
 
 FLINK_VERSION_LATEST: 1.8-SNAPSHOT
@@ -20,30 +19,6 @@ FLINK_ISSUES_URL: https://issues.apache.org/jira/browse/FLINK
 FLINK_GITHUB_URL: https://github.com/apache/flink
 FLINK_CONTRIBUTORS_URL: https://cwiki.apache.org/confluence/display/FLINK/List+of+contributors
 FLINK_GITHUB_REPO_NAME: flink
-
-optional_components:
-  -
-    name: "Avro"
-    category: "SQL Formats"
-    scala_dependent: false
-    versions:
-      -
-        version: "1.7.2"
-        id: 172-sql-format-avro
-        url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-avro/1.7.2/flink-avro-1.7.2.jar
-        asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-avro/1.7.2/flink-avro-1.7.2.jar.asc
-        sha_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-avro/1.7.2/flink-avro-1.7.2.jar.sha1
-  -
-    name: "Json"
-    category: "SQL Formats"
-    scala_dependent: false
-    versions:
-      -
-        version: "1.7.2"
-        id: 172-sql-format-json
-        url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar
-        asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar.asc
-        sha_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar.sha1
 
 # Example for scala dependent component:
 #  -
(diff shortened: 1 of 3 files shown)


aljoscha:
I updated the PR, it's now split into two commits, one for updating the Downloads page and one for adding 1.8 and removing 1.5.6. I think it makes sense to do this in one PR because adding 1.8 is the motivation for the change in the Download page, and because you can see what it looks like for Flink 1.8 this way.

I restructured the config file so that everything for a given Flink release is now in one place, see also the commit message on that commit. Additional components that are tied to a Flink release are in that structure, additional components that are independent of a Flink release are in an independent list and in a separate section on the Downloads page.

screenshot 2019-03-01 at 15 12 05
screenshot 2019-03-01 at 15 12 22

downloads.zh.md
```diff
@@ -18,120 +18,134 @@ $( document ).ready(function() {
...
-    <!-- overrride margin/padding as the boxes otherwise overlap in subtle ways -->
-    <h4 style="margin-top: 0px; padding-top: 0px;"><span class="glyphicon glyphicon-download" aria-hidden="true"></span> <strong>{{ source_release.name }}</strong> Source Release</h4>
-  </a>
-   (<a href="{{ source_release.asc_url }}">asc</a>, <a href="{{ source_release.sha512_url }}">sha512</a>)
-</div>
-{% endfor %}
+Apache Flink® {{ site.FLINK_VERSION_STABLE }} is our latest stable release.
+
+An Apache Hadoop installation is [not required](faq.html#how-does-flink-relate-to-the-hadoop-stack) to use Apache Flink.
+For users that use Flink without any Hadoop components, we recommend the release without bundled Hadoop libraries.
```
tweise: That's confusing. Remove since we don't have the option with bundled Hadoop any longer?

aljoscha: Will change that 👍

_config.yml
```diff
@@ -21,30 +20,6 @@ FLINK_GITHUB_URL: https://github.com/apache/flink
...
-    scala_dependent: false
-    versions:
-      -
-        version: "1.7.2"
-        id: 172-sql-format-json
-        url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar
-        asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar.asc
-        sha_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-json/1.7.2/flink-json-1.7.2.jar.sha1
-
 # Example for scala dependent component:
```
zentol: This example is now out-dated.

aljoscha: Should I just remove the example? We currently don't have any scala dependent optional components.

tzulitai:
I think it's against ASF regulations for project release download pages to not include a link to source releases:
http://www.apache.org/dev/release-download-pages.html#links

> The download page must include a link to the source distribution. It may include links to binary distributions.

_config.yml
zentol: I'm fine with removing it since the alternative binaries section can serve as a guideline.

_config.yml
```diff
@@ -76,166 +51,223 @@ optional_components:
...
+          url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar
+          asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar.asc
+          sha_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar.sha1
+        -
+          name: "Pre-bundled Hadoop 2.8.3"
+          category: "Pre-bundled Hadoop"
+          scala_dependent: false
+          id: 180-bundled-hadoop-283
+          url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.8.3-1.8.0/flink-shaded-hadoop2-uber-2.8.3-1.8.0.jar
+          asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.8.3-1.8.0/flink-shaded-hadoop2-uber-2.8.3-1.8.0.jar.asc
```
sunjincheng121: Can not download, we got `404 - Path /org/apache/flink/flink-shaded-hadoop2-uber/2.8.3-1.8.0/flink-shaded-hadoop2-uber-2.8.3-1.8.0.jar not found in group repository "Public Repositories" [id=public].`

_config.yml
```diff
@@ -76,166 +51,223 @@ optional_components:
...
+          id: 180-bundled-hadoop-275
+          url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar
+          asc_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar.asc
+          sha_url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.7.5-1.8.0/flink-shaded-hadoop2-uber-2.7.5-1.8.0.jar.sha1
+        -
+          name: "Pre-bundled Hadoop 2.8.3"
+          category: "Pre-bundled Hadoop"
+          scala_dependent: false
+          id: 180-bundled-hadoop-283
+          url: https://repository.apache.org/content/groups/public/org/apache/flink/flink-shaded-hadoop2-uber/2.8.3-1.8.0/flink-shaded-hadoop2-uber-2.8.3-1.8.0.jar
```
sunjincheng121: Can not download, we got `404 - Path /org/apache/flink/flink-shaded-hadoop2-uber/2.8.3-1.8.0/flink-shaded-hadoop2-uber-2.8.3-1.8.0.jar not found in group repository "Public Repositories" [id=public].`

zentol: of course you can'T download them, they aren't released yet.

PR #187, opened 2019-03-13
[FLINK-11893] [Project Website] Update ecosystem to encourage sharing of the connectors

becketqin:
Based on the discussion in the mailing list, we would like to encourage the connector authors to share their connectors with the community, even the connectors are not in the Flink repo.

This ticket updates the ecosystem page on the website to reflect the change. It also adds ecosystem page back to the navigation bar.

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -63,6 +63,9 @@
               </ul>
             </li>
 
+            <!-- Ecosystem -->
+            <li{% if page.url contains '/ecosystem.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/ecosystem.html">{{ site.data.i18n[page.language].ecosystem }}</a></li>
+
             <!-- getting help -->
             <li{% if page.url contains '/gettinghelp.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/gettinghelp.html">{{ site.data.i18n[page.language].getting_help }}</a></li>
 
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -63,11 +63,11 @@
               </ul>
             </li>
 
-            <!-- Ecosystem -->
-            <li{% if page.url contains '/ecosystem.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/ecosystem.html">{{ site.data.i18n[page.language].ecosystem }}</a></li>
-
             <!-- getting help -->
             <li{% if page.url contains '/gettinghelp.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/gettinghelp.html">{{ site.data.i18n[page.language].getting_help }}</a></li>
+
+            <!-- Ecosystem -->
+            <li{% if page.url contains '/ecosystem.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/ecosystem.html">{{ site.data.i18n[page.language].ecosystem }}</a></li>
 
             <!-- Blog -->
             <li{% if page.url contains '/blog/' or page.url contains '/news/' or page.url contains '/features/' %} class="active"{% endif %}><a href="{{ site.baseurl }}/blog/"><b>{{ site.data.i18n[page.language].flink_blog }}</b></a></li>
--- a/content/ecosystem.html
+++ b/content/ecosystem.html
@@ -98,6 +98,9 @@
               </ul>
             </li>
 
+            <!-- Ecosystem -->
+            <li class="active"><a href="/ecosystem.html">Ecosystem</a></li>
+
             <!-- getting help -->
             <li><a href="/gettinghelp.html">Getting Help</a></li>
 
@@ -170,19 +173,105 @@ many other data processing projects and frameworks.
 
 <p>Connectors provide code for interfacing with various third-party systems.</p>
 
-<p>Currently these systems are supported:</p>
+<p>
+  The following table lists the connectors that are available in Flink ecosystem.
+  Connector authors are welcome to add your connector to this list via a Github
+  Pull Request. See <a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/improve-website.html"> Improve the Website</a>
+  for more details.
+</p>
 
-<ul>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/kafka.html" target="_blank">Apache Kafka</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/elasticsearch.html" target="_blank">Elasticsearch 1.x / 2.x / 5.x / 6.x</a> (sink)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/filesystem_sink.html" target="_blank">HDFS</a> (sink)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/rabbitmq.html" target="_blank">RabbitMQ</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/kinesis.html" target="_blank">Amazon Kinesis Streams</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/twitter.html" target="_blank">Twitter</a> (source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/nifi.html" target="_blank">Apache NiFi</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/cassandra.html" target="_blank">Apache Cassandra</a> (sink)</li>
-  <li><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a> (sink)</li>
-</ul>
+<table class="table table-bordered">
+  <tr>
+    <th>System Name</th>
+    <th>Connector Type</th>
+    <th>Location</th>
+    <th>License</th>
+    <th>Last released</th>
+    <th>Compatible Flink Versions</th>
+    <th>Maintained By</th>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/kafka.html" target="_blank">Apache Kafka</a></td>
+    <td>sink/source</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/elasticsearch.html" target="_blank">Elasticsearch 1.x / 2.x / 5.x / 6.x</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/filesystem_sink.html" target="_blank">HDFS</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/rabbitmq.html" target="_blank">RabbitMQ</a></td>
+    <td>sink/source</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/kinesis.html" target="_blank">Amazon Kinesis Streams</a></td>
+    <td>sink/source</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/twitter.html" target="_blank">Twitter</a></td>
+    <td>source</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/nifi.html" target="_blank">Apache NiFi</a></td>
+    <td>sink/source</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/cassandra.html" target="_blank">Apache Cassandra</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+</table>
 
 <p>To run an application using one of these connectors, additional third party
 components are usually required to be installed and launched, e.g., the servers
--- a/content/ecosystem.html
+++ b/content/ecosystem.html
@@ -98,9 +98,6 @@
               </ul>
             </li>
 
-            <!-- Ecosystem -->
-            <li class="active"><a href="/ecosystem.html">Ecosystem</a></li>
-
             <!-- getting help -->
             <li><a href="/gettinghelp.html">Getting Help</a></li>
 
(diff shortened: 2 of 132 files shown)


fhueske:
Hi @becketqin, thanks for the PR!

Can you remove all changes generated by the build script?
That will make the PR easier to review.

Thank you, Fabian

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
```
rmetzger: This needs to be "Bahir Repository"

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
+    <td><a href="{{site.docs-stable}}/dev/connectors/cassandra.html" target="_blank">Apache Cassandra</a></td>
+    <td>sink</td>
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a></td>
```
rmetzger: I think it would be nicer of Redis, Flume and ActiveMQ would all get their own row in the table.
There are now more connectors available in Apache Bahir: https://github.com/apache/bahir-flink

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
-  <li><a href="{{site.docs-stable}}/dev/connectors/filesystem_sink.html" target="_blank">HDFS</a> (sink)</li>
-  <li><a href="{{site.docs-stable}}/dev/connectors/rabbitmq.html" target="_blank">RabbitMQ</a> (sink/source)</li>
-  <li><a href="{{site.docs-stable}}/dev/connectors/kinesis.html" target="_blank">Amazon Kinesis Streams</a> (sink/source)</li>
-  <li><a href="{{site.docs-stable}}/dev/connectors/twitter.html" target="_blank">Twitter</a> (source)</li>
-  <li><a href="{{site.docs-stable}}/dev/connectors/nifi.html" target="_blank">Apache NiFi</a> (sink/source)</li>
-  <li><a href="{{site.docs-stable}}/dev/connectors/cassandra.html" target="_blank">Apache Cassandra</a> (sink)</li>
-  <li><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a> (sink)</li>
-</ul>
+<p>
+  The following table lists the connectors that are available in Flink ecosystem.
```
rmetzger:
```suggestion
  The following table lists the connectors that are available in the Flink ecosystem.
```

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
+  Connector authors are welcome to add your connector to this list via a Github
```
rmetzger:
```suggestion
  Connector authors are welcome to add their connector to this list via a Github
```

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
+  Pull Request. See <a href="{{site.docs-stable}}/improve-website.html"> Improve the Website</a>
```
rmetzger:
```suggestion
  Pull Request. See <a href="{{site.docs-stable}}/improve-website.html">Improve the Website</a>
```

ecosystem.md
```diff
@@ -11,19 +11,105 @@ many other data processing projects and frameworks.
...
+    <td>Flink Repo</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+</table>
 
 To run an application using one of these connectors, additional third party
 components are usually required to be installed and launched, e.g., the servers
```
rmetzger: Can you add "https://github.com/TouK/nussknacker/" to the list of 3rd party projects?

rmetzger:
Thank you @becketqin ! I left some comments. Once these are resolved, I'm +1 to merge.

_includes/navbar.html
```diff
@@ -63,6 +63,9 @@
               </ul>
             </li>
 
+            <!-- Ecosystem -->
+            <li{% if page.url contains '/ecosystem.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/ecosystem.html">{{ site.data.i18n[page.language].ecosystem }}</a></li>
```
fhueske: I would move the Ecosystem link below "Getting Help".

content/ecosystem.html
```diff
@@ -170,19 +173,105 @@ <h2 id="connectors">Connectors</h2>
...
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/filesystem_sink.html" target="_blank">HDFS</a> (sink)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/rabbitmq.html" target="_blank">RabbitMQ</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/kinesis.html" target="_blank">Amazon Kinesis Streams</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/twitter.html" target="_blank">Twitter</a> (source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/nifi.html" target="_blank">Apache NiFi</a> (sink/source)</li>
-  <li><a href="https://ci.apache.org/projects/flink/flink-docs-release-1.7/dev/connectors/cassandra.html" target="_blank">Apache Cassandra</a> (sink)</li>
-  <li><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a> (sink)</li>
-</ul>
+<p>
```
fhueske: I think it would be good to make a pass over the third party projects and remove all that are not in a reasonable shape. For example the Cascading integration should be removed and there are probably others that are not maintained. Linking them does not provide any benefit and might even give the feeling that the page is not maintained.

fhueske (commented):
I've added a few comments as well.
Thanks, Fabian

becketqin:
@fhueske @rmetzger Thanks a lot for the review. I updated the PR based on your comments. I removed the changes to the html files from build script. I will do a rebuild and add them back before the PR is merged.

ecosystem.md
```diff
@@ -11,19 +11,141 @@ many other data processing projects and frameworks.
...
-  <li><a href="https://github.com/apache/bahir-flink" target="_blank">Redis, Flume, and ActiveMQ (via Apache Bahir)</a> (sink)</li>
-</ul>
+<p>
+  The following table lists the connectors that are available in the Flink ecosystem.
+  Connector authors are welcome to add their connector to this list via a Github
+  Pull Request. See <a href="{{site.docs-stable}}/improve-website.html">Improve the Website</a>
+  for more details.
+</p>
+
+<table class="table table-bordered">
```
fhueske: Does it make sense to add another column with a link to the documentation?

rmetzger:
+1 to merge from my point of view

ecosystem.md
```diff
@@ -11,19 +11,141 @@ many other data processing projects and frameworks.
...
+  </tr>
+  <tr>
+    <td><a href="http://bahir.apache.org/docs/flink/1.0/flink-streaming-redis/" target="_blank">Redis</a></td>
+    <td>sink/source</td>
+    <td><a href="https://github.com/apache/bahir-flink">Apache Bahir</a></td>
+    <td>Apache 2.0</td>
+    <td>05/24/2017</td>
+    <td>1.7.x</td>
+    <td>Apache Bahir</td>
+  </tr>
```
fhueske: Apache Pulsar features Flink connectors (https://github.com/apache/pulsar/tree/master/pulsar-flink/src/main/java/org/apache/flink). Would be great to have it listed here as well.

fhueske: Can of course be added later as well.

fhueske (commented):
Thanks for the update @becketqin!

I made another pass and left a few comments. 
Another issue that we should consider is how to keep this list up to date when a new Flink version is released. For the Flink connectors, we could replace the version string by the version variable such that it is automatically incremented when a new version is released. For the third party connectors, we might need to do this manually. 

Best, Fabian

becketqin (commented):
@fhueske Thanks for the comments. I replied with my thoughts. Please let me know what do you think.

ecosystem.md
```diff
@@ -11,19 +11,141 @@ many other data processing projects and frameworks.
...
+<p>
+  The following table lists the connectors that are available in the Flink ecosystem.
+  Connector authors are welcome to add their connector to this list via a Github
+  Pull Request. See <a href="{{site.docs-stable}}/improve-website.html">Improve the Website</a>
+  for more details.
+</p>
+
+<table class="table table-bordered">
+  <tr>
+    <th>System Name</th>
```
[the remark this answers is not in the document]
fhueske: Yes, that would be great!

ecosystem.md
```diff
@@ -11,19 +11,141 @@ many other data processing projects and frameworks.
...
+    <td><a href="{{site.docs-stable}}/dev/connectors/elasticsearch.html" target="_blank">Elasticsearch 1.x / 2.x / 5.x / 6.x</a></td>
+    <td>sink</td>
+    <td>Apache Flink</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>1.7.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
+    <td><a href="{{site.docs-stable}}/dev/connectors/filesystem_sink.html" target="_blank">HDFS</a></td>
```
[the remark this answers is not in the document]
fhueske: We could replace `HDFS` by `File Systems (HDFS, S3, and others)` and make others a link.

Regarding the encodings, I think we should list them in a separate section. Putting them here would clutter the connector list, but I think it is important to show that Flink supports different encodings like Parquet and ORC. Having a file system connector does not help a lot if the encoding that a user needs is not supported.

becketqin (commented):
@fhueske I just updated the patch. One thing to mention is that it looks we do not have a JDBC connector page. So I left the JDBC connector without a link. I'll create a followup ticket to add a JDBC page then link it from here.

ecosystem.md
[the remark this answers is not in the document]
becketqin: It is absolutely necessary to mention the supported encoding information. But this seems something specific to particular file systems, instead of applicable to all connectors. So would it be better to have the encoding information in the individual pages of those specific systems?

becketqin:
@fhueske Ping. Do you have time to take another look?

rmetzger:
I'd like to get this in asap, otherwise, the real community packages website will be done before this is merged :)

becketqin:
@rmetzger Yeah... I agree. @fhueske could you take look? Maybe we can check in the current update and make further modifications in separate patches if needed?

fhueske:
Hi, sorry @becketqin, I forgot about this PR.
I'll have a look later today.

Thanks, Fabian

ecosystem.md
```diff
@@ -50,12 +307,6 @@ See also Jim Dowling's [Flink Forward talk](http://www.slideshare.net/FlinkForwa
 [Apache Mahout](https://mahout.apache.org/) is a machine learning library that will feature Flink as an execution engine soon.
 Check out Sebastian Schelter's [Flink Forward talk](http://www.slideshare.net/FlinkForward/sebastian-schelter-distributed-machine-learing-with-the-samsara-dsl) about Mahout-Samsara DSL.
 
-**Cascading**
-
-[Cascading](http://www.cascading.org/cascading-flink/) enables a user to build complex workflows easily on Flink and other execution engines.
-[Cascading on Flink](https://github.com/dataArtisans/cascading-flink) is built by [dataArtisans](http://data-artisans.com/) and [Driven, Inc](http://www.driven.io/).
-See Fabian Hueske's [Flink Forward talk](http://www.slideshare.net/FlinkForward/fabian-hueske-training-cascading-on-flink) for more details.
-
 **Apache Beam**
```
fhueske: I think we should we order projects alphabetically.

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <th>Connector Type</th>
+    <th>Location</th>
+    <th>License</th>
+    <th>Last Released</th>
+    <th>Available For</th>
+    <th>Compatible Flink Versions</th>
+    <th>Maintained By</th>
+  </tr>
+  <tr>
+    <td><a href="{{site.docs-stable}}/dev/connectors/kafka.html" target="_blank">Apache Kafka</a></td>
```
fhueske: should we link here to the Apache Kafka website? 
We can add the link to Flink's documentation to the "Available For" cell. This would allow us to link to the docs of the different APIs (the docs of the Table API Kafka connector are different from the DataStream API Kafka connector docs).

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+  Pull Request. See <a href="{{site.docs-stable}}/improve-website.html">Improve the Website</a>
+  for more details.
+</p>
+
+### Message Queue
+<table class="table table-bordered">
+  <tr>
+    <th>System Name</th>
+    <th>Connector Type</th>
+    <th>Location</th>
```
fhueske: Do we need a "Location" and a "Maintained By" column? It seems these are duplicate.

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <td>sink</td>
+    <td><a href="https://github.com/apache/bahir-flink">Apache Bahir</a></td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>DataStream</td>
+    <td>{{site.stable}}.x</td>
+    <td>Apache Bahir</td>
+  </tr>
+  <tr>
+    <td><a href="http://bahir.apache.org/docs/flink/current/flink-streaming-influxdb/" target="_blank">InfluxDB</a></td>
```
fhueske: duplicate entry

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+  </tr>
+  <tr>
+    <td>JDBC</td>
+    <td>sink/source</td>
+    <td>Apache Flink</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>Table</td>
+    <td>{{site.stable}}.x</td>
+    <td>Apache Bahir</td>
```
fhueske: JDBC is maintained by the Flink community

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+  </tr>
+  <tr>
+    <td>JDBC</td>
+    <td>sink/source</td>
+    <td>Apache Flink</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>Table</td>
```
fhueske: We have sources and sinks for the DataSet API as well

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <th>Available For</th>
+    <th>Compatible Flink Versions</th>
+    <th>Maintained By</th>
+  </tr>
+  <tr>
+    <td><a href="{{site.docs-stable}}/dev/connectors/kafka.html" target="_blank">Apache Kafka</a></td>
+    <td>sink/source</td>
+    <td>Apache Flink</td>
+    <td>Apache 2.0</td>
+    <td></td>
```
fhueske: should we leave this cell empty? 
For us it is clear that this is the last Flink release, but visitors might be confused about the missing date.
Should we add a variable with the latest Flink release here?

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <td><a href="{{site.docs-stable}}/dev/connectors/streamfile_sink.html" target="_blank">Others File Systems (S3, others)</a></td>
+    <td>sink</td>
+    <td>Apache Flink</td>
+    <td>Apache 2.0</td>
+    <td></td>
+    <td>DataSet / Table</td>
+    <td>{{site.stable}}.x</td>
+    <td>Apache Flink</td>
+  </tr>
+  <tr>
```
fhueske: We have file system source connectors for DataStream and DataSet

fhueske: We have file system connectors for DataSet API

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <td><a href="{{site.docs-stable}}/dev/connectors/streamfile_sink.html" target="_blank">Others File Systems (S3, others)</a></td>
```
fhueske: I would list all file systems under the same entry. We should not treat HDFS different that other FSs, IMO.

ecosystem.md
```diff
@@ -11,19 +11,276 @@ many other data processing projects and frameworks.
...
+    <td><a href="http://bahir.apache.org/docs/flink/1.0/flink-streaming-redis/" target="_blank">Redis</a></td>
+    <td>sink/source</td>
+    <td><a href="https://github.com/apache/bahir-flink">Apache Bahir</a></td>
+    <td>Apache 2.0</td>
+    <td>05/24/2017</td>
+    <td>DataStream</td>
+    <td>{{site.stable}}.x</td>
+    <td>Apache Bahir</td>
+  </tr>
+</table>
```
fhueske: We have an HBase source connector for DataSet and Table APIs.

ecosystem.md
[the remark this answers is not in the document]
fhueske: The supported encodings are independent of the file system. However, they depend on the connector implementation (Flink API). For example, we support ORC as source format for DataSet and Table API (and probably also DataStream via continuous file source) but not as sink format.

fhueske (commented):
Hi @becketqin,

I've left a few comments.

Best, Fabian

becketqin (commented):
@fhueske Thanks for the review. I made some modifications and replied to some of the comments. In general, I would consider this page as a fast reference to the connectors and ecosystem projects, rather than providing too much details.

becketqin:
@rmetzger @fhueske One thing not quite clear to me is the `compatible Flink version` section. As far as I understand, anything compatible with `Flink X.Y` should also be compatible with `Flink X.Z` as long as `Z` >= `Y`. Is that correct? If that is the case, I'll check at which Flink version was each connector added, and update the `Compatible Flink Versions` accordingly.

ecosystem.md
fhueske: I don't think that adding links in "Available For" would mean that we need to have different pages/links for the different APIs. If the links are the same, we can simply link to the same page, IMO.

However, in most cases we have different documentation pages for the different APIs. For example the file system connectors:
* Streaming 
   * [Streaming File Sink](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/connectors/streamfile_sink.html)
   * [Monitoring File Source](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/datastream_api.html#data-sources) (for streaming)
* Batch
  * [Source](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/batch/#data-sources)
  * [Sink](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/batch/#data-sinks)
* Table
  * [File System Source](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/table/connect.html#file-system-connector)
  * [ORC File Sink](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/table/connect.html#csvtablesink)

ecosystem.md
fhueske: I'm fine keeping both. 
At this point it just looked duplicate to me.

ecosystem.md
fhueske: The interfaces are [org/apache/flink/api/common/io/FileInputFormat](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/io/FileInputFormat.java) and [org/apache/flink/api/common/io/FileOutputFormat](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/io/FileOutputFormat.java).

Concrete implementations depend on the format (like ORC, CSV, etc.). There are also wrappers for Hadoop InputFormat implementation like Parquet, SequenceFile, etc.

ecosystem.md
fhueske: Flink's `FileSystem` interface hides manu of the differences of the different supported file systems (HDFS, S3, ...) from the connector implementations.

fhueske:
Thanks for the update @becketqin. I agree that this page should be a quick reference to more detailed information. However, I also think that it should be complete (at least with respect to build-in connectors). One of the main problems is the large number of possible combinations (APIs, source/sink, encodings, etc.).

Regarding your question: This is usually true. However, we sometimes also drop support for outdated connectors or older versions of other systems.

rmetzger:
> @rmetzger @fhueske One thing not quite clear to me is the `compatible Flink version` section. As far as I understand, anything compatible with `Flink X.Y` should also be compatible with `Flink X.Z` as long as `Z` >= `Y`. Is that correct? If that is the case, I'll check at which Flink version was each connector added, and update the `Compatible Flink Versions` accordingly.

How about removing that information from the page?
I would prefer to have something about our ecosystem on the Flink website, instead of discussing forever here.

rmetzger:
Once this is merged, let's add https://issues.apache.org/jira/browse/FLINK-12783

fhueske:
I'm fine leaving the compatibility version out. 

As I said before, I think the information about available build-in connectors should be complete as this page is likely to be considered as source-of-truth wrt. connector availability and it would be a shame if users would be discouraged because they cannot find info about an available connector.

becketqin:
@rmetzger Thanks for pushing this. I'll remove the compatible versions column.

> I'm fine leaving the compatibility version out.
> 
> As I said before, I think the information about available build-in connectors should be complete as this page is likely to be considered as source-of-truth wrt. connector availability and it would be a shame if users would be discouraged because they cannot find info about an available connector.

Hi @fhueske , in general I agree that this page should serve as a good connector availability reference for the users. However, it is unclear to me what is considered `complete`. For example, some security mechanism may not be supported, or we only have append sink while user needs an upsert sink, or some source connector may support timestamp while some others do not. Is the availability information of these connectors incomplete? To the users who needs that, such connector is not available. But including them here will simply explode the page and make it difficult to maintain.

So I think we need to draw a line between "available" and "fully meet user requirements". 
Personally speaking, I would take connector type, supported API and compatible versions as necessary for users to consider a connector to be available. And other information, such as supported encoding, supported security mechanisms, whether it is an upsert or append sink, belong to whether the connector "fully meets user requirements". We can keep them in the individual pages of each connectors.

Thoughts?

MartijnVisser:
Completed via https://issues.apache.org/jira/browse/FLINK-22922

PR #197, opened 2019-04-10
[FLINK-12157] Use maven central links

zentol:
Updates maven artifact links since we're not allowed to point to `repository.apache.org`.
I got this URL from https://central.sonatype.org/pages/consumers.html.

Someone else needs to regenerate the docs since I'm currently not able to do so.


aljoscha:
merged 👌 😊

PR #198, opened 2019-04-10
Fix link of language switch is broken in blog page

wuchong:
The link of language switch in blog page should link to homepage of Chinese version, otherwise, the URL is 404 not found.


fhueske:
Thanks for the fix @wuchong!
Merging

PR #204, opened 2019-04-30
Add multi-language support section(python table api) to the roadmap

sunjincheng121:
Add multi-language support section(python table API) to the roadmap.

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -154,7 +154,7 @@ the catalogs ([FLINK-10232](https://issues.apache.org/jira/browse/FLINK-10232)).
 There is a broad effort to integrate Flink with the Hive Ecosystem, including
 metastore and Hive UDF support [FLINK-10556](https://issues.apache.org/jira/browse/FLINK-10556).
 
-There also a big effort to support Python for Table API [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API).
+There is also a big effort to support Python for Table API [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API).
 We will divide the work into following stages:
 
 - Translate Python Table API queries without UDFs to Java and run them completely in Java for the first step.
--- a/roadmap.md
+++ b/roadmap.md
@@ -154,6 +154,13 @@ the catalogs ([FLINK-10232](https://issues.apache.org/jira/browse/FLINK-10232)).
 There is a broad effort to integrate Flink with the Hive Ecosystem, including
 metastore and Hive UDF support [FLINK-10556](https://issues.apache.org/jira/browse/FLINK-10556).
 
+There also a big effort to support Python for Table API [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API).
+We will divide the work into following stages:
+
+- Translate Python Table API queries without UDFs to Java and run them completely in Java for the first step.
+- Add support for User-defined functions(Scalar Function/Table Function/Aggregate Function) in the second step.
+- Integrating Pandas as the final effort, i.e., functions in Pandas can be used in Python Table API directly.
+
 # Connectors & Formats
 
 Support for additional connectors and formats is a continuous process.
--- a/roadmap.zh.md
+++ b/roadmap.zh.md
@@ -100,6 +100,13 @@ Flink Web UI 正在移植到更新的框架中并获得其他功能并更好地�
 
 还有一个巨大的工作是将 Flink 与 Hive 生态系统集成。包括 Metastore 和 Hive UDF 支持 [FLINK-10556](https://issues.apache.org/jira/browse/FLINK-10556)。
 
+社区在Table API支持Python方面也做出了很多的努力 [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API)。
+我们将工作分为以下几个阶段:
+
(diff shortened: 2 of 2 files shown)


roadmap.md
```diff
@@ -72,6 +72,13 @@ On the runtime level, the streaming operators are extended to also support the d
 patterns required for some batch operations ([discussion thread](https://lists.apache.org/thread.html/cb1633d10d17b0c639c3d59b2283e9e01ecda3e54ba860073c124878@%3Cdev.flink.apache.org%3E)).
 This is also groundwork for features like efficient [side inputs](https://cwiki.apache.org/confluence/display/FLINK/FLIP-17+Side+Inputs+for+DataStream+API).
 
+# Multi-language support
```
fhueske: I'd make this more specific: `Python support for Table API`

roadmap.md
```diff
@@ -72,6 +72,13 @@ On the runtime level, the streaming operators are extended to also support the d
...
+
+At the Flink's API level, we have already support DataStream API/DataSet API/Table API&SQL, and both DataSet API and 
+DataStream API already provides python API(flink-python and flink-stream-python, but they may be deprecated later). 
+To cater to the current large number of analytics users, we plan to redesign the Flink Python API and support
```
fhueske: I would not call this a "redesign" of the Python API but a new Python Table API.

roadmap.md
```diff
@@ -72,6 +72,13 @@ On the runtime level, the streaming operators are extended to also support the d
 patterns required for some batch operations ([discussion thread](https://lists.apache.org/thread.html/cb1633d10d17b0c639c3d59b2283e9e01ecda3e54ba860073c124878@%3Cdev.flink.apache.org%3E)).
 This is also groundwork for features like efficient [side inputs](https://cwiki.apache.org/confluence/display/FLINK/FLIP-17+Side+Inputs+for+DataStream+API).
 
+# Multi-language support
+
+At the Flink's API level, we have already support DataStream API/DataSet API/Table API&SQL, and both DataSet API and 
+DataStream API already provides python API(flink-python and flink-stream-python, but they may be deprecated later). 
+To cater to the current large number of analytics users, we plan to redesign the Flink Python API and support 
+Table API first since Table API will become the first-class citizen in Flink. [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API)
```
fhueske: Should we add a bit more details like first step is to translate Python Table API queries without UDFs to Java and run them completely in Java and add support for UDFs in a second step?

roadmap.md
fhueske: I would move this down Ecosystem and Connectors as the other efforts are more "mature" and made good progress already.

roadmap.md
```diff
@@ -72,6 +72,13 @@ On the runtime level, the streaming operators are extended to also support the d
 patterns required for some batch operations ([discussion thread](https://lists.apache.org/thread.html/cb1633d10d17b0c639c3d59b2283e9e01ecda3e54ba860073c124878@%3Cdev.flink.apache.org%3E)).
 This is also groundwork for features like efficient [side inputs](https://cwiki.apache.org/confluence/display/FLINK/FLIP-17+Side+Inputs+for+DataStream+API).
 
+# Multi-language support
+
+At the Flink's API level, we have already support DataStream API/DataSet API/Table API&SQL, and both DataSet API and
```
bowenli86: The first sentence feels more like an intro to background, and doesn't cut to the key point right away. How about moving it to the end and making "To cater to ...." the opening?

wuchong:
Hi @sunjincheng121 , could you also add a same paragraph to `roadmap.zh.md` or translate it in `roadmap.zh.md` ?

sunjincheng121:
Thanks for your valuable comments @fhueske @bowenli86 @wuchong!
I agree we can move the Python Table API section into the Ecosystem, in this way, we do not need to introduce more future plans for multi-language support (there is no good progress)  and put a simple and straightforward description of the progress of the Python Table API directly.

BTW: The reason I added the multi-language part is that I want to give users a clear understanding of Flink's subsequent multi-language support. Python is the first step in multi-language support. But I think we can also add that information in the Ecosystem if we need in the future. 

I have updated the PR according to your comments.  I appreciate if you can have another look!

Best,
Jincheng

wuchong:
Thanks for the PR @sunjincheng121 . The Chinese part looks good to me.

wuchong:
One suggestion for the roadmap, might not be the scope of this PR.

Shall we add a `latest-update-time` to the top of Roadmap page ? So that users can know this is a up-to-date Roadmap.

sunjincheng121:
Thanks for your review. @wuchong 
Do you have time to have another look for the update? @fhueske @bowenli86 
Thanks,
Jincheng

roadmap.md
```diff
@@ -154,6 +154,14 @@ the catalogs ([FLINK-10232](https://issues.apache.org/jira/browse/FLINK-10232)).
 There is a broad effort to integrate Flink with the Hive Ecosystem, including
 metastore and Hive UDF support [FLINK-10556](https://issues.apache.org/jira/browse/FLINK-10556).
 
+There also a big effort to support Python for Table API [FLIP-38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API).
```
bowenli86: There 'is' also ...

sunjincheng121: Must we add `is`, just double check?

bowenli86: No offense. This is just a minor grammar thing that I thought you made a typo :)  'there' used like this case is always followed by an auxiliary verb, like "there is..., there are..., there will be..." etc

sunjincheng121:
Thanks all of your review! @fhueske @wuchong @bowenli86 
I'll merge this PR if there no other comments.

sunjincheng121:
Merging...

PR #205, opened 2019-05-01
Updating gradle to work with latest gradle

chethanuk:
When Gradle version is more than 4.4, the ` gradle clean shadowJar` will produce an error: "> No value has been specified for property 'mainClassName'." So updated the gradle-quickstart to fix the error with latest gradle

I got many errors with old quickstart:
```
* What went wrong:
A problem was found with the configuration of task ':shadowJar'.
> No value has been specified for property 'mainClassName'.
```

So, I updated the gradle-quickstrat

### DIFF
--- a/q/gradle-quickstart.sh
+++ b/q/gradle-quickstart.sh
@@ -113,7 +113,8 @@ plugins {
     id 'java'
     id 'application'
     // shadow plugin to produce fat JARs
-    id 'com.github.johnrengelman.shadow' version '2.0.4'
+    // If gradle version is greater than 4.4, then this will produce error: No value has been specified for property 'mainClassName'.
+    // id 'com.github.johnrengelman.shadow' version '2.0.4'
 }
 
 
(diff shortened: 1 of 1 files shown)


NicoK (changes requested):
Thanks @ChethanUK for your contribution and sorry for the delay. A few details before we continue:

1. Please create a [JIRA](http://issues.apache.org/jira/browse/FLINK) ticket for Flink (and adapt this PR's title)
2. Please also create a pull request for `docs/dev/projectsetup/java_api_quickstart.md` in the Flink repository (this needs to be kept in sync with the quickstart script)
3. Were you able to setup the gradle wrapper? In that case, you should have had at least a working build file, even without your PR.
4. Why did you not specify a plugin version that you require? https://imperceptiblethoughts.com/shadow/getting-started/ describes two way of applying the plugin but you kind of chose a mix. While at it, you could update to the latest version
5. Have you verified the packing is still correct (only classes in jar which are supposed to be there? compare with mvn!), even if you include things like a Kafka connector or the Table API?

MartijnVisser:
Closing this PR as this has been resolved via https://github.com/apache/flink-web/pull/504

PR #226, opened 2019-07-15
Fix wrong subscribe URL in community page

Yikun:
The subscribe URL should be `*listname*-subscribe@flink.apache.org`, but current is `subscribe-*listname*@flink.apache.org`.

This patch try to fix it and change it as the correct link.


rmetzger:
Good catch! 
Thanks a lot for fixing this. I'll merge the PR.

PR #229, opened 2019-07-18
Correct typo in roadmap.md

TisonKun:
"an" should be "and"


rmetzger:
Thanks for the fix. Merging!

PR #234, opened 2019-07-23
 [FLINK-13342] Update master documentation link to say "master"

aljoscha:
At least one user was confused by the fact that during releases the "master" documentation does point to the wrong snapshot, see description in the issue.


tzulitai:
LGTM, +1

aljoscha:
Thanks Gordon!

PR #241, opened 2019-08-11
Add builds mailing list and update nabble archive for user-zh mailing list

wuchong:
As discussed in the dev mailing list, we have setup a builds mailing list to deliver travis build notifications, so that the community can get the notification result in time. 

In this commit, we also update nabble archive for user-zh mailing list.


rmetzger:
Thanks a lot!
Merging ...

PR #266, opened 2019-09-18
.htaccess file added

realabbas:
# .htaccess file added

**For better browser leverage caching**

Currently, the page speed is 41%. One of the factors is the lack of **browser caching**

image

**Zero Browser Caching**

Screenshot (1131)

Adding the .htaccess file will enable the browser caching and hence will increase the speed.

Full Website Report Link: https://gtmetrix.com/reports/flink.apache.org/1blCRYfs

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,12 +1,19 @@
-# redirect rules to prevent old links from breaking
-RewriteEngine On
+<IfModule mod_expires.c>
+  ExpiresActive On
 
-RewriteCond %{HTTPS} off
-RewriteRule ^(.*)$ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
+  # Images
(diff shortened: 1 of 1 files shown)


NicoK (changes requested):
I'm actually a bit sceptical for enabling caching here because it could happen that we have to change an image which we published, e.g. for fixing something in a blog post. Maybe a less aggressive 1day caching is fine, similarly for css and js. Not sure how much that helps then (and how much is already covered by modern browsers) but that sounds like something that would leave some flexibility for us.
This change should probably be discussed on the flink-dev ml though.

In general though, you need to keep the old rewrite rules.

MartijnVisser:
After the migration to Hugo, I've ran this check once more. Score is now:

image

Therefore closing this PR

PR #272, opened 2019-09-25
[FLINK-14213] Replace link to "Local Setup Tutorial" by link to "Getting Started Overview".

fhueske:
We are improving the starting experience for new Flink users by adding new code walkthroughs and Docker-based playgrounds. However, the new material is not easy to find from the Flink website.

To make the material more visible, I'm replacing the prominent *"Tutorials"* link on the Flink website which points to the [local setup tutorial](https://ci.apache.org/projects/flink/flink-docs-release-1.9/getting-started/tutorials/local_setup.html) by a *"Getting Started"* link that points to the [Getting Started Overview](https://ci.apache.org/projects/flink/flink-docs-release-1.9/getting-started/) page.

### DIFF
--- a/_data/i18n.yml
+++ b/_data/i18n.yml
@@ -7,7 +7,7 @@ en:
     powered_by: Powered By
     faq: FAQ
     downloads: Downloads
-    tutorials: Tutorials
+    getting_started: Getting Started
     documentation: Documentation
     getting_help: Getting Help
     flink_blog: Flink Blog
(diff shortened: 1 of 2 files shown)


sjwiesman:
@fhueske I'm moving a bunch of files around in apache/flink#10384 that includes removing the local setup tutorial. There are redirects in place so this won't break the website but we should be sure to get this PR quickly after the 1.10 release.

fhueske:
Thanks for the comment @sjwiesman. I think we could also merge the PR now, because it replaces the link to local setup by the new getting started overview page which is available on the 1.9 docs as well

sjwiesman:
+1 to merge

fhueske:
Thanks for the review @sjwiesman.
Merged!

PR #282, opened 2019-11-18
add link to flink-packages.org

rmetzger:
This is how the change looks rendered: 
Screenshot 2019-11-14 at 12 53 35


zentol:
Looks a bit lost doesn't it? I'd prefer a dedicated section on the downloads or maybe community(???) page, explaining that this is a page for third-party components that are not release by apache.

zentol:
At the same time I'm unsure whether we are even allowed to link to third-party sites from the download page.

rmetzger:
I'm not sure if people would look at the download page for Flink extensions, if they are not really intending to download the software, but rather exploring it.
The community page would be an option, but I fear that it will not get enough attention.

One major downside of my current approach is indeed that I'm not adding any text that explains the packages site.

becketqin:
I was wondering whether adding the website to ecosystem web page would be better. Ecosystem page is pretty intuitive for users to click when looking for something around Flink. And we can also put more descriptions there for the website.

StephanEwen:
I agree with Robert - the community page would be too hidden. I kind of like it in the sidenav. Maybe instead of the "Ecosystem" link?

About "Ecosystem" - my guess is that the community packages replace the static "Ecosystem" page. Its almost the same, except that "packages" is more structured and user-editable.

rmetzger:
Thanks Stephan. I updated the PR to put the flink-packages.org link instead of the "Ecosystem" link.

becketqin:
@rmetzger Makes sense to me.

README.md
```diff
@@ -2,4 +2,4 @@
 
 This repository contains the Flink website: https://flink.apache.org/.
 
-You find instructions for this repository here: https://flink.apache.org/contributing/improve-website.html.
\ No newline at end of file
+You find instructions for this repository here: https://flink.apache.org/contributing/improve-website.html.
```
zentol: It appears git is trying to play spot-the-difference 😕

_includes/navbar.html
```diff
@@ -83,9 +83,11 @@
             <!-- Blog -->
             <li{% if page.url contains '/blog/' or page.url contains '/news/' or page.url contains '/features/' %} class="active"{% endif %}><a href="{{ site.baseurl }}/blog/"><b>{{ site.data.i18n[page.language].flink_blog }}</b></a></li>
 
-            <!-- Ecosystem -->
-            <li{% if page.url contains '/ecosystem.html' %} class="active"{% endif %}><a href="{{ baseurl_i18n }}/ecosystem.html">{{ site.data.i18n[page.language].ecosystem }}</a></li>
 
+            <!-- Flink-packages -->
+            <li>
+              <a href="https://flink-packages.org" target="_blank">{{ site.data.i18n[page.language].packages }} <small><span class="glyphicon glyphicon-new-window"></span></small></a>
```
zentol: Wondering whether `Ecosystem` isn't actually more expressive than `flink-packages.org`

ecosystem.md
```diff
@@ -1,106 +1,5 @@
 ---
 title: "Ecosystem"
----
-<br>
-Apache Flink supports a broad ecosystem and works seamlessly with
```
zentol: Why not keep this page and frontpage link, and just update it to point to flink-packages.org + explanation as to what it is?
Are we just trying to reduce overhead since f-p.o already contains a description on the main page?

_includes/navbar.html
rmetzger: The current link name makes it pretty clear that the page is external.

ecosystem.md
rmetzger: I also thought about introducing an intermediate page, but users would have to kick through one level of indirection to get to the page.
I fear that people will not find the link to flink-package.org.

rmetzger:
@zentol are you ok to merge this as is?

PR #292, opened 2020-01-17
[hotfix] Correct payer/recepient terminology

afedulov:
A small correction is needed to correct the terminology. Payee and beneficiary are the same entity, payee is changed to payer.


fhueske:
Merged

PR #297, opened 2020-01-31
[FLINK-15823] Sync download page translation for Chinese

hequn8128:
Adjust the Chinese version of `downloads.md` for the changes made in https://issues.apache.org/jira/browse/FLINK-15800

Furthermore, the version display under `Release Notes`  is wrong for the Chinese download page, so change `site.FLINK_VERSION_STABLE_SHORT` to `flink_release.version_short`.

### DIFF
--- a/content/zh/downloads.html
+++ b/content/zh/downloads.html
@@ -194,22 +194,9 @@ $( document ).ready(function() {
 
 <div class="page-toc">
 <ul id="markdown-toc">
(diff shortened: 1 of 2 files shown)


hequn8128:
Will merge this directly since it is a simple PR which tries to sync [FLINK-15800](https://issues.apache.org/jira/browse/FLINK-15800)

PR #302, opened 2020-02-04
Update downloads.md for Flink 1.10.0

GJL:


_config.yml
```diff
@@ -94,137 +139,6 @@ flink_releases:
           url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-json/1.9.2/flink-json-1.9.2.jar
           asc_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-json/1.9.2/flink-json-1.9.2.jar.asc
           sha_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-json/1.9.2/flink-json-1.9.2.jar.sha1
-  -
-      version_short: 1.8
```
zentol: we could move this removal into a separate commit

GJL:
Thanks for the review.

_config.yml
GJL: done

PR #306, opened 2020-02-14
Update expired links in the Powered By page.

morsapaes:
The use cases for King and Telefónica DE point to links that aren't valid anymore.


fhueske:
Thanks for fixing the links @morsapaes!

I merged the PR

PR #311, opened 2020-03-03
[hotfix] Remove mention of icla requirement on how to contrib guide

rmetzger:
I would like to remove this part of the flink website, as it regularly leads to first time contributors submitting the document through the asf secretary, which is not the purpose of the section.


StephanEwen:
Good idea!

+1 from my side

rmetzger:
Thanks for the review. Merging!

PR #327, opened 2020-04-14
Fix format error of md file. 

gk0916:
This error caused the list to not be printed correctly when the md file was converted to an html file


gk0916:
fix to list format.

PR #329, opened 2020-04-18
Add Apache Flink release 1.9.3

dianfu:
Adds release announcement and download link for 1.9.3
Also note that FLINK-16954 doesn't appear in the release note as it's a fix of FLINK-11767 which was introduced in 1.9.3.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -103,48 +103,48 @@ flink_releases:
   -
       version_short: 1.9
       binary_release:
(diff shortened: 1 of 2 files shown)


dianfu:
@carp84 @uce Thanks a lot for the review. I will update the post date and release date before merging this PR.

dianfu:
Merged via af2d8de132b73f0dae6fd88912fe1c92202f3efe

PR #333, opened 2020-05-01
[FLINK-17490] Add training page

alpinegizmo:
Now that the documentation has a training section, it would be good to help folks find it by promoting it from the project website.

This adds training.md and training.zh.md, and adds a Training entry to the site navigation.

image

### DIFF
--- a/training.md
+++ b/training.md
@@ -1,0 +1,107 @@
+---
+title: "Training"
+---
+
+The Apache Flink community maintains a self-paced training course that contains
+a set of lessons and hands-on exercises. This step-by-step introduction to Flink focuses
+on learning how to use the DataStream API to meet the needs of common, real-world use cases.
+
+This training covers the fundamentals of Flink, including:
+
+<div class="row">
+    <div class="col-sm-4">
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <span class="glyphicon glyphicon-blackboard"></span>  <b>Intro to Flink</b>
+            </div>
+            <div class="panel-body">
+                <ul style="font-size: small;">
+                    <li>Batch vs. Streaming</li>
+                    <li>Parallel Dataflows</li>
+                    <li>State, Time, and Snapshots</li>
+                </ul>
+            </div>
+        </div>
+    </div>
+    <div class="col-sm-4">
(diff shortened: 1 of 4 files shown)


alpinegizmo:
I've created https://issues.apache.org/jira/browse/FLINK-17491 for the Chinese translation.

morsapaes:
Hey, David! Thanks a lot for doing the whole integration of the training — it's a super valuable resource that indeed deserves more attention.

What about linking this from the "Getting Started" dropdown, instead? I'm afraid we're starting to accumulate quite a lot of disparate links in the navigation bar. Once the discussion to revamp the landing page kicks in, then we could make sure that the self-paced training is properly highlighted in there.

XBaith:
> Hey, David! Thanks a lot for doing the whole integration of the training — it's a super valuable resource that indeed deserves more attention.
> 
> What about linking this from the "Getting Started" dropdown, instead? I'm afraid we're starting to accumulate quite a lot of disparate links in the navigation bar. Once the discussion to revamp the landing page kicks in, then we could make sure that the self-paced training is properly highlighted in there.

Good idea.I also think that putting "Training" in "Get Started" will help users quickly and systematically learn with Flink

alpinegizmo:
@morsapaes @XBaith I've changed it, based on your feedback. 

image

morsapaes:
Thanks, David! I think this makes it more intuitive to find. 👌

XBaith:
Thank you @alpinegizmo .

training.md
```diff
@@ -0,0 +1,107 @@
+---
+title: "Training Course"
+---
+
```
NicoK, line 4: Other pages seem to add this to titles; we should be consistent
```suggestion

<hr/>

```

training.zh.md
```diff
@@ -0,0 +1,107 @@
+---
+title: "Training Course"
+---
+
```
NicoK, line 4: Other pages seem to add this to titles; we should be consistent
```suggestion

<hr/>

```

NicoK (changes requested):
I like the page +1

(seems like titles are always separated by `<hr>` ... once this is added, we can go ahead)

NicoK (approved):

PR #355, opened 2020-07-06
Update website for Flink 1.11.0

pnowojski:


_config.yml
```diff
@@ -55,6 +55,51 @@ FLINK_GITHUB_REPO_NAME: flink-statefun
...
+          url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-avro/1.11.0/flink-avro-1.11.0.jar
+          asc_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-avro/1.11.0/flink-avro-1.11.0.jar.asc
+          sha_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-avro/1.11.0/flink-avro-1.11.0.jar.sha1
+        -
+          name: "CSV SQL Format"
+          category: "SQL Formats"
+          scala_dependent: false
+          id: 1110-sql-format-csv
+          url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-csv/1.11.0/flink-csv-1.11.0.jar
+          asc_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-csv/1.11.0/flink-csv-1.11.0.jar.asc
```
zentol: json/csv formats are now bundled in lib/, so a separate download seems unnecessary.

zentol (approved):
1.9 should be removed from the downloads page, but beyond that (and the one minor redundant download listing) this looks correct.

_config.yml
pnowojski: Shouldn't this be mentioned in the release notes?

zentol: Yes. It appears this was missed in https://issues.apache.org/jira/browse/FLINK-18173.

pnowojski: I've removed those links and I'm adding release notes:
https://github.com/apache/flink/pull/12828

pnowojski:
Thanks for the review. I've removed 1.9 download links as well. Can you take a look again @zentol ?

zentol (approved):
+1

PR #357, opened 2020-07-09
[hotfix] update training links to point to stable docs

alpinegizmo:
Now that Flink 1.11 has been released, the training is available in the stable docs. That's where these links should go.


sjwiesman:
LGTM, merging.

PR #374, opened 2020-08-31
[hotfix] Introduce parameter to make future posts visible in local run

XComp:
build.sh parameters as part of the run. Additionally, a new parameter
got introduced to build.sh that extends -p to also enable posts
happening in the future.


docker-run.sh
```diff
@@ -16,10 +16,4 @@
 # limitations under the License.
```
rmetzger: The file misses the shebang.

rmetzger: Maybe calling the file `docker-build.sh` makes more sense, since it's calling the `build.sh` script, and this script wraps it in docker?

rmetzger: @sjwiesman introduced the Makefile like 2 weeks ago. I don't care if it is a makefile or not, but I was wondering if he had a strong reason to add a Makefile.

sjwiesman: No strong reason, I like makefiles :shrug: 

Either solution is fine with me

XComp: I just found it better fitting since we only do parameter passing in the end. I found that the intend of make does not include parameter passing. Is this a correct conclusion? I didn't want to include a third target to support another parameter.

rmetzger (approved):
Change looks good.

PR #399, opened 2020-12-08
Add Apache Flink release 1.11.3

xintongsong:
Add release 1.11.3. Dates are to be updated.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -87,32 +87,32 @@ flink_releases:
   -
       version_short: "1.11"
       binary_release:
-          name: "Apache Flink 1.11.2"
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -160,7 +160,12 @@ component_releases:
 
 release_archive:
     flink:
-      - version_short: "1.12"
+      -
+        version_short: "1.11"
```
dianfu, line 167: All the 1.11 series should be grouped together.

xintongsong: True. I thought they were ordered by release date. Will fix this. Thanks.

tzulitai:
+1, LGTM

PR #400, opened 2020-12-16
[hotfix] Add missed JOIN keyword for sql example in the announcement

leonardBang:
Add missed AS keyword for sql example in the announcement.

* checked in plan test


twalthr (approved):
LGTM

PR #408, opened 2021-01-05
Add security page for Flink

rmetzger:


security.md
```diff
@@ -0,0 +1,74 @@
...
+Apache Flink is a framework for executing user-supplied code in clusters. Users can submit jar files to Flink processes, which will be executed unconditionally, without any attempts to limit what code can run. Starting other processes, establishing network connections or accessing and modifing local files is possible.
+
+Historically, we've received numerous remote code execution vulnerability reports, which we had to reject, as this is by design.
+
+**We strongly discourage users to expose Flink processes to the public internet**. Within company networks or "cloud" accounts, we recommend restricting the access to a Flink cluster via appropriate means.
+
+
+### I found a vulnerability in Flink, how do I report it?
+
+Thanks a lot for lookiing into the security of Apache Flink! We appreciate reports improving the security of Flink. We accept vulnerability reports through the [Apache Security Team](http://www.apache.org/security/), via their private email address [security@apache.org](mailto:security@apache.org).
```
zentol:
```suggestion
Thanks a lot for looking into the security of Apache Flink! We appreciate reports improving the security of Flink. We accept vulnerability reports through the [Apache Security Team](http://www.apache.org/security/), via their private email address [security@apache.org](mailto:security@apache.org).
```

security.zh.md
```diff
@@ -0,0 +1,74 @@
...
+Apache Flink is a framework for executing user-supplied code in clusters. Users can submit jar files to Flink processes, which will be executed unconditionally, without any attempts to limit what code can run. Starting other processes, establishing network connections or accessing and modifing local files is possible.
+
+Historically, we've received numerous remote code execution vulnerability reports, which we had to reject, as this is by design.
+
+**We strongly discourage users to expose Flink processes to the public internet**. Within company networks or "cloud" accounts, we recommend restricting the access to a Flink cluster via appropriate means.
+
+
+### I found a vulnerability in Flink, how do I report it?
+
+Thanks a lot for lookiing into the security of Apache Flink! We appreciate reports improving the security of Flink. We accept vulnerability reports through the [Apache Security Team](http://www.apache.org/security/), via their private email address [security@apache.org](mailto:security@apache.org).
```
zentol:
```suggestion
Thanks a lot for looking into the security of Apache Flink! We appreciate reports improving the security of Flink. We accept vulnerability reports through the [Apache Security Team](http://www.apache.org/security/), via their private email address [security@apache.org](mailto:security@apache.org).
```

_includes/navbar.html
```diff
@@ -177,7 +177,9 @@
 
               <a class="smalllinks" href="https://www.apache.org/licenses/" target="_blank">License</a> <small><span class="glyphicon glyphicon-new-window"></span></small>
 
-              <a class="smalllinks" href="https://www.apache.org/security/" target="_blank">Security</a> <small><span class="glyphicon glyphicon-new-window"></span></small>
+              <a class="smalllinks" href="{{ baseurl_i18n }}/security.html">Flink Security</a>
```
zentol: This kinda ruins the current layout of the sidebar.

Screenshot_2021-01-06 Apache Flink Stateful Computations over Data Streams

Maybe we could throw out some other item in the sidebar instead, like the plan visualizer.

security.md
```diff
@@ -0,0 +1,74 @@
...
+	<tr>
+		<td>
+			<a href="https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-1960">CVE-2020-1960</a>
+		</td>
+		<td>
+			1.1.0 to 1.1.5, 1.2.0 to 1.2.1, 1.3.0 to 1.3.3, 1.4.0 to 1.4.2, 1.5.0 to 1.5.6, 1.6.0 to 1.6.4, 1.7.0 to 1.7.2, 1.8.0 to 1.8.3, 1.9.0 to 1.9.2, 1.10.0
+		</td>
+		<td>
+			<a href="https://github.com/apache/flink/commit/804ae70024bf8be7c0c7093d02addb080c318662">Fixed in commit 804ae70024bf8be7c0c7093d02addb080c318662</a> <br>
+			Users are advised to upgrade to Flink 1.9.3 or 1.10.1 or later versions.
```
zentol: This not only excludes some information from the actual advisory we published (at the very least changing the configuration should be listed as an option), but for users upgrading to 1.9.3 the commit hash is also wrong, and thus of questionable value (also applies to other entries). Maybe just omit the commit hash entirely.

security.md
```diff
@@ -0,0 +1,74 @@
...
+		</td>
+	</tr>
+</table>
+
+
+## Frequently Asked Questions
+
+### During a security analysis of Flink, I noticed that Flink allows for remote code execution, is this an issue?
+
+Apache Flink is a framework for executing user-supplied code in clusters. Users can submit jar files to Flink processes, which will be executed unconditionally, without any attempts to limit what code can run. Starting other processes, establishing network connections or accessing and modifing local files is possible.
```
zentol:
```suggestion
Apache Flink is a framework for executing user-supplied code in clusters. Users can submit code to Flink processes, which will be executed unconditionally, without any attempts to limit what code can run. Starting other processes, establishing network connections or accessing and modifying local files is possible.
```
Given the existence of python/sql jobs, "jar files" seems outdated. Maybe just call it "code".

security.md
```diff
@@ -0,0 +1,74 @@
...
+
+Historically, we've received numerous remote code execution vulnerability reports, which we had to reject, as this is by design.
+
+**We strongly discourage users to expose Flink processes to the public internet**. Within company networks or "cloud" accounts, we recommend restricting the access to a Flink cluster via appropriate means.
```
zentol:
```suggestion
**We strongly discourage users to expose Flink processes to the public internet**. Within company networks or "cloud" accounts, we recommend restricting access to a Flink cluster via appropriate means.
```

zentol (changes requested):

_includes/navbar.html
rmetzger: Is this better?

image

zentol: it is _better_, but...like the current layout for the apache-mandated stuff is nigh perfect :/

rmetzger: I vaguely remember that the plan visualizer is kinda popular (people complained when we removed it) .. that's why I want to keep it.

How about this? 
image

zentol: I think that's the better alternative at this time. (although I'm worried that over time the apache stuff gets pushed down more and more)

rmetzger: I agree that we need to be careful. Thanks for confirming.

security.md
rmetzger: I removed the commit sha.

zentol (approved):
+1, although I'm not sure whether we should list the commit hash for the other CVEs given that we explicitly did not include them in the advisory.

rmetzger:
Thanks for the approval. I included the commits because it would be very difficult to find the commits for people wanting to backport the fixes.

PR #409, opened 2021-01-08
[FLINK-20832] Deliver bootstrap css and js ourselves

wuchong:
Delivering the required files ourselves, instead of using a CDN to fix the unstable https://maxcdn.bootstrapcdn.com in China.


rmetzger (approved):
Change looks good.
Tested it locally.

Will merge now.

PR #411, opened 2021-01-08
[FLINK-20900][docs] Added guidelines for documenting commands.

XComp:
As part of refactoring the deployment documentation we came up with some agreements that we could add to the docs guidelines to get a more consistent look & feel:

long parameter names for commands
one parameter per line
parameter line indented
command preceded by `$`
Options

### DIFF
--- a/contributing/docs-style.md
+++ b/contributing/docs-style.md
@@ -463,6 +463,32 @@ Or using plain HTML:
     might not be obvious from reading it. Use comments to clarify
     implementation details and to describe the expected output.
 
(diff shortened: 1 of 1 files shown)


rmetzger (approved):
Thanks a lot! 
these are good additions in my opinion.

I'm leaving this open for a bit longer to give others the opportunity to raise concerns, if there are any.

sjwiesman:
+1 thanks for taking the initiative on this!

rmetzger:
Thx for taking another look. Merging it now

PR #419, opened 2021-02-20
Missing ending doublequote for defaultFlinkVersion.

leewallen:
The "Run the quickstart script" gradle quickstart example fails because of a missing ending quotation mark.


rkhachatryan (approved):
Thanks for the fix, LGTM.

PR #423, opened 2021-02-26
Fix broken link to Hadoop YARN integration

gvauvert:


zentol:
Closing in favor of #425; I think that removing the paragraph entirely is the better option; it doesn't really make sense for YARN to be specifically listed on that page.
Thank you for trying to fix it, and apologies we haven't seen this PR sooner.

PR #425, opened 2021-03-10
[FLINK-21723] Remove link to Hadoop integration from downloads page

tillrohrmann:
Yarn is no longer the only resource provider Flink users can use. Hence,
we should no longer directly point from the downloads page to this one
resource provider implementation.

### DIFF
--- a/content/downloads.html
+++ b/content/downloads.html
@@ -236,10 +236,6 @@ $( document ).ready(function() {
 
 <p>Apache Flink® 1.12.2 is our latest stable release.</p>
 
-<p>If you plan to use Apache Flink together with Apache Hadoop (run Flink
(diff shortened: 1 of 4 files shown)


zentol (approved):
+1, but it would be good to figure out where all the whitespace changes in the .md files came from. (If you modified them in IntelliJ, is the Save Actions plugin possible modifying them?)

tillrohrmann:
Thanks for the review @zentol. It must have been my Atom editor which added these whitespaces. I will revert the whitespace changes and then I'll merge this PR.

tillrohrmann:
Hmm, when looking at the whitespace changes, I noticed that we have a couple of lines which end with a whitespace and then with a `\n` character. I think this is what Atom has corrected. Hence I would think that this change is actually correct. I will separate it into a a hotfix commit though.

PR #426, opened 2021-03-15
Update Roadmap

StephanEwen:
This updates the roadmap of the website, as discusses in this thread.

https://lists.apache.org/x/thread.html/ra8a4e311835c361260bf32f5e6f59a979d88b23e61a8379e89280cbe@%3Cdev.flink.apache.org%3E

The relevant document is [roadmap.md](https://github.com/apache/flink-web/pull/426/files#diff-51abe96bc493857fdfba599c6b70e3a6c09200cf29e130d15b599fef538e32df).

## Follow-ups

The Chinese language version of the page needs to be translated / adjusted as well.

## Question regarding style:

There are two variants of the Feature Radar in this PR. Reviewers, please let us know which option you think we should pick in the end.

  - A variant that is more Flink-color scheme and Font: https://raw.githubusercontent.com/StephanEwen/flink-web/ebbb89410962eafd96d18ff4becba007b581bc5d/content/img/flink_feature_radar.svg
    

  - A variant with a different color scheme that might be easier to read: https://raw.githubusercontent.com/StephanEwen/flink-web/ebbb89410962eafd96d18ff4becba007b581bc5d/content/img/flink_feature_radar_2.svg


rmetzger (approved):
Thanks a lot for this PR!

+1 to merge.

I like the 2nd version of the radar a bit more, because its easier to read.

knaufk (commented):
A little bit late, but what do you think about adding the "History Server" as a "Approaching End of Life" component. Is was added in Flink 1.3 and has not received development attention since then. Alternatively, it could also be an Beta/MVP feature, but it might be weird to have an MVP feature that was introducedin Flink 1.3 and never moved beyond that. cc @tillrohrmann @zentol

tillrohrmann:
Hard to tell how actively the history server is used by our users. Given that we almost never see complaints about it, I think it is not used very often. Hence, we might think about transitioning it out.

StephanEwen:
@tillrohrmann @knaufk Could you do an assessment with a Mailing List Poll / Twitter Poll / etc.?
Would be good to get a few more data points, if we don't seem to know well what the state is.

StephanEwen:
Thanks everyone. I would merge the second version then.
Once there is a conclusion about the History Server, I'll update the chart.

PR #429, opened 2021-04-07
[FLINK-22004] Translate Flink Roadmap to Chinese

curcur:
1. Translate Flink Roadmap to Chinese
2. Feature Radar (Figure) may also need to be translated
3. Already built locally.


curcur:
should pull against apache:asf-site instead of apache:master,

close this one.

PR #430, opened 2021-04-08
[FLINK-22004] Update Flink Roadmap Chinese Version 

curcur:
1. Translate Flink Roadmap to Chinese
2. Already built and tested locally
3. update the feature radar picture as well


StephanEwen:
Thanks a lot for the translation and the feature radar. Nice work!
Will merge this now...

The website rebuild seems to have dropped some pages, but I'll just rebuild the website during the merge.

StephanEwen:
Merged it.

Turns out the website rebuilding was completely correct, the html diff just looked surprising initially.
(I assume that is because it included changes from previous commits that were not propagated from md to html, yet.)

PR #431, opened 2021-04-08
[FLINK-22116] Setup .asf.yaml for publishing

zentol:
Publishing websites from git without an .asf.yaml will not work from June onwards.


knaufk (approved):
LGTM based on https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features.

PR #446, opened 2021-05-10
Added release 1.12.4

AHeise:


_config.yml
```diff
@@ -159,7 +159,7 @@ release_archive:
         release_date: 2021-04-30
       - 
         version_short: "1.12"
-        version_long: 1.12.3
+        version_long: 1.12.4
```
zentol, line 162: This is incorrect; you need to add a new entry

zentol (changes requested):

_config.yml
AHeise: Okay, fixed. Could you please recheck?

zentol (approved):

PR #456, opened 2021-07-05
[FLINK-23257] Update docker-build.sh

zentol:
We need Ruby 2.6 to build the docs. The newer image does not ship with bundler being installed by default, so we have to do it manually.


sjwiesman (approved):
Ugh, we really need to get the site off Jekyll.

PR #457, opened 2021-07-05
[FLINK-23258] Make incremental builds explicit and log a warning

zentol:


sjwiesman:
What do you think about having the message explicitly direct committers to not use incremental builds when regenerating the site for publication?

zentol:
Sounds good. +1

sjwiesman (approved):
+1

PR #467, opened 2021-09-16
Updating Flink Feature Radar

MartijnVisser:
Updated the Flink Feature Radar:
* Python DataStream API from MVP to Beta
* Expanded Unified Sink API with Kafka, renamed FileSink to File
* Moved Change-Data-Capture API and Connectors from **Beta** to **Production Ready & Evolving**
* Added Hybrid Source to **Beta**
* Moved Unified Source API from **Beta** to **Production Ready & Evolving** and expanded with Pulsar
* Moved Kubernetes-based HA (ZK-alternative) from **Beta** to **Production Ready & Evolving**
* Removed Pulsar Source & Sink (since these specifically are not bundled with Flink but externally and they are being replaced because new Pulsar Source is included via Unified Source API and the new sink will be built based on Unified Sink API
* Added Default as part of new Scheduler category to **Stable**
* Added Adaptive Scheduler / Reactive Mode as part of Scheduler category to **Beta**
* Added Source Function & Sink Function to **Approaching End-Of-Life** to mark that old source and sink interfaces will be deprecated in the near future
* Moved File Source & Sink from **Stable** to **Approaching End-Of-Life** (since they use the old source and sink interfaces)
* Moved Kafka Source & Sink from **Stable** to **Approaching End-Of-Life** (since they use the old source and sink interfaces)
* Removed Legacy SQL Query Engine from **Deprecated** since it's removed in 1.14
* Removed Mesos from **Deprecated** since it's removed in 1.14


StephanEwen:
Thanks, this looks good. Just two questions:

  1. In the "Phasing Out" part, should we rename "Kafka Source & Sink" to "Legacy Kafka Source & Sink", and similarly for File S&S? To avoid confusion, because the new Kafka Source / Sinks are not super easy to see (being under Unified Source/Sink API).
  We might even want to have a separate entry for Kafka and File, like the other connectors, just because those connectors are so essential and I expect many users will look for them.
  
  2. Is the Active Kubernetes integration still evolving? I don't recall many changes in 1.14 anymore, so is it ready to be labeled as stable?

MartijnVisser:
1. I've renamed the Phasing Out ones to Legacy for both Kafka and File. I've also renamed the new ones to Kafka, File, Pulsar (via Unified Source API) and Kafka, File (via Unified Sink API) to put the emphasis on the connector. I think that when the legacy ones are removed, we'll break them up into its own connector
2. I've checked that when with the team that's maintaining it and they wanted to keep it at Production Ready & Evolving for now

StephanEwen:
Thanks, this looks good. Mergin this...

PR #475, opened 2021-10-14
[hotfix] Add Analytics GA4 implementation to be prepared for migration

MartijnVisser:
Google has a new implementation to track website behaviour, called GA4. This will replace the existing UA implementations and is not compatible with the current version of tracking behaviour. 

This PR implements the GA4 tracking next to the UA tracking, so that at a later stage the UA tracking can be removed. 

Commit a928385c0f8ccc66bdf8f29ae8e968a1b1754563 makes sure that the old implementation anonymises IP addresses by default. The new implementation already does this by default. See https://support.google.com/analytics/answer/2763052?hl=en


MartijnVisser:
Superseded by https://github.com/apache/flink-web/pull/502

PR #490, opened 2021-12-14
Add flink-shaded 15.0

zentol:


XComp (changes requested):
It looks like a manual build of the website is missing which regenerates the resources in the `contents/` subfolder?

There are still `1.14` references in `contents/downloads.html`

zentol:
> It looks like a manual build of the website is missing 

Well yes, I only add those before the merge to save time.

> There are still 1.14 references in contents/downloads.html

1.14 reference are possibly for Flink itself?

XComp (approved):
Yeah, I double-checked. Not sure what crossed my mind when doing the initial review. ...should be fine

PR #493, opened 2021-12-28
Add flink-ml 2.0.0

gaoyunhaii:
This PR updates the documents to include the downloads information for the upcoming flink-ml 2.0.0

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -23,6 +23,12 @@ FLINK_STATEFUN_STABLE_SHORT: "3.1"
 FLINK_STATEFUN_GITHUB_URL: https://github.com/apache/flink-statefun
 FLINK_STATEFUN_GITHUB_REPO_NAME: flink-statefun
 
+FLINK_ML_VERSION_STABLE: 2.0.0
+FLINK_ML_STABLE_SHORT: "2.0"
+
+FLINK_ML_GITHUB_URL: https://github.com/apache/flink-ml
+FLINK_ML_GITHUB_REPO_NAME: flink-ml
+
 # Example for scala dependent component:
 #  -
 #    name: "Prometheus MetricReporter"
@@ -150,6 +156,17 @@ flink_statefun_releases:
           url: "https://www.apache.org/dyn/closer.lua/flink/flink-statefun-3.1.1/flink-statefun-3.1.1-src.tgz"
           asc_url: "https://downloads.apache.org/flink/flink-statefun-3.1.1/flink-statefun-3.1.1-src.tgz.asc"
           sha512_url: "https://downloads.apache.org/flink/flink-statefun-3.1.1/flink-statefun-3.1.1-src.tgz.sha512"
+
+flink_ml_releases:
+  -
+      version_short: "2.0"
+      source_release:
+          name: "Apache Flink ML 2.0.0"
+          id: "200-ml-download-source"
+          flink_version: "1.14.0"
+          url: "https://www.apache.org/dyn/closer.lua/flink/flink-ml-2.0.0/flink-ml-2.0.0-src.tgz"
+          asc_url: "https://downloads.apache.org/flink/flink-ml-2.0.0/flink-ml-2.0.0-src.tgz.asc"
+          sha512_url: "https://downloads.apache.org/flink/flink-ml-2.0.0/flink-ml-2.0.0-src.tgz.sha512"
 
 component_releases:
   -
@@ -561,6 +578,11 @@ release_archive:
       -
         version: 1.0
         release_date: 2017-07-27
+
+    flink_ml:
+      - version_short: 2.0
+        version_long: 2.0.0
+        release_date: 2021-01-07
 
 # Version numbers used in the text for stable and snapshot versions,
 # e.g. "Documentation for {{ site.stable }".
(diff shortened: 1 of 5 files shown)


downloads.md
```diff
@@ -200,6 +200,20 @@ You can add the following dependencies to your `pom.xml` to include Apache Flink
...
+
+```xml
+<dependency>
+  <groupId>org.apache.flink</groupId>
+  <artifactId>flink-ml-uber_2.12</artifactId>
+  <version>{{ site.FLINK_ML_VERSION_STABLE }}</version>
+</dependency>
+```
+
+The `flink-ml-uber_2.12` is a whole set of dependencies required for both using and developing algorithms. To only use existing algorithms, users could also only depend on `flink-ml-lib_2.12` and to only develop new algorithms, user could also only depend on `flink-ml-core` and `flink-ml-iteration` (if the algorithm is iterable).
```
lindong28: This paragraph seems a bit difficult to read.

Could we use the following words:

`flink-ml-uber_2.12` provides all the dependencies needed to use Flink ML. Alternatively, advanced users could use the following artifacts to import a minimal set of Flink ML dependencies for their target use-cases:

- Use artifact `flink-ml-core` in order to develop custom ML algorithms.
- Use artifacts `flink-ml-core` and `flink-ml-iteration` in order to develop custom ML algorithms which require iteration.
- Use artifact `flink-ml-lib_2.12` in order to use the off-the-shelf ML algorithms from Flink ML.

downloads.md
```diff
@@ -200,6 +200,24 @@ You can add the following dependencies to your `pom.xml` to include Apache Flink
...
+
+```xml
+<dependency>
+  <groupId>org.apache.flink</groupId>
+  <artifactId>flink-ml-uber_2.12</artifactId>
+  <version>{{ site.FLINK_ML_VERSION_STABLE }}</version>
+</dependency>
+```
+
+flink-ml-uber_2.12 provides all the dependencies needed to use Flink ML. Alternatively, advanced users could use the following artifacts to import a minimal set of Flink ML dependencies for their target use-cases:
```
zhipeng93: It seems that flink-ml-uber_2.12 only includes `flink-ml-core` and `flink-ml-iteration` [1]. Do you mean to include all other dependencies in flink-ml?

[1] https://github.com/apache/flink-ml/blob/master/flink-ml-uber/pom.xml

_config.yml
```diff
@@ -23,6 +23,12 @@ FLINK_STATEFUN_STABLE_SHORT: "3.1"
 FLINK_STATEFUN_GITHUB_URL: https://github.com/apache/flink-statefun
 FLINK_STATEFUN_GITHUB_REPO_NAME: flink-statefun
 
+FLINK_ML_VERSION_STABLE: 2.0.0
```
lindong28, line 26: If we search `state` in this file, there seems to be some content related to statefun which do not have similar Flink ML counterpart. Could you help add these? Thanks!

tillrohrmann (commented):
Thanks for creating this PR @gaoyunhaii. I think this PR is missing the rebuild commit.

_config.yml
gaoyunhaii: Very thanks for the review~ I complete the `flink_ml_releases` and the corresponding part. The doc part could be updated after the flink-ml doc get online~

downloads.md
gaoyunhaii: Hi @zhipeng93 I think you are right, I removed dependency of `flink-ml-uber` currently and use directly the direct dependencies . We could re-enabled the `uber` jar in the next version.

gaoyunhaii:
Very thanks @tillrohrmann for the review! I add the rebuild commit now~

_config.yml
```diff
@@ -151,6 +157,17 @@ flink_statefun_releases:
           asc_url: "https://downloads.apache.org/flink/flink-statefun-3.1.1/flink-statefun-3.1.1-src.tgz.asc"
           sha512_url: "https://downloads.apache.org/flink/flink-statefun-3.1.1/flink-statefun-3.1.1-src.tgz.sha512"
 
+flink_ml_releases:
+  -
+      version_short: "2.0"
+      source_release:
+          name: "Apache Flink ML 2.0.0"
+          id: "200-ml-download-source"
+          flink_version: "1.14.2"
```
dianfu: Should this be 1.14.0? ( https://github.com/apache/flink-ml/blob/release-2.0/pom.xml#L73)

gaoyunhaii: Yes, indeed, I changed the version~

_config.yml
```diff
@@ -562,6 +579,11 @@ release_archive:
         version: 1.0
         release_date: 2017-07-27
 
+    flink_ml:
+      - version_short: 2.0
+        version_long: 2.0.0
+        release_date: 2021-12-30
```
lindong28: Could we change this to `2022-01-07` to be more accurate?

lindong28 (commented):
Could we also update `./content/zh/downloads.html`? Thanks!

gaoyunhaii:
@lindong28 I added the translation~

lindong28:
@gaoyunhaii Thanks for the update! I have checked both `http://0.0.0.0:4000/downloads.html` and `http://0.0.0.0:4000/zh/downloads.html` locally and the pages LGTM.

dianfu (approved):

PR #500, opened 2022-01-21
[FLINK-25748] tidy up source code section on community page

knaufk:


alpinegizmo (approved):
I think this is fine, as far as it goes. There are some other parts of this page that need an update as well, but I can do that in a follow-on PR later.

PR #504, opened 2022-02-02
[FLINK-25897] Update gradle quickstart to gradle 7.3.3

matriv:
Update gradle quickstart to be working with gradle `7.3.3`

### DIFF
--- a/q/gradle-quickstart.sh
+++ b/q/gradle-quickstart.sh
@@ -42,7 +42,10 @@ defaultProjectName="quickstart"
 defaultOrganization="org.myorg.quickstart"
 defaultVersion="0.1-SNAPSHOT"
 defaultFlinkVersion="${1:-1.14.4}"
-defaultScalaBinaryVersion="${2:-2.11}"
+# flink-docs-master/docs/dev/datastream/project-configuration/#gradle
+# passes the scala version prefixed with a _, e.g.: _2.12
+scalaBinaryVersionFromCmdArg="${2/_/}"
+defaultScalaBinaryVersion="${scalaBinaryVersionFromCmdArg:-2.12}"
 
 echo "This script creates a Flink project using Java and Gradle."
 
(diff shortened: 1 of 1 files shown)


q/gradle-quickstart.sh
```diff
@@ -232,6 +223,13 @@ else
   exit 1
 fi
 
+# create gradle dir with gradle wrapper
+mkdir -p gradle/wrapper
+cat > gradle/wrapper/gradle-wrapper.properties <<EOF
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-all.zip
+EOF
```
dmvk: this is not really useful unless we also provide `./gradlew` scripts for actually running the wrapper

dmvk (approved):
LGTM overall, I have one concern about usability of the attached gradle wrapper, please take a look

Thanks for the PR Marios 👍

q/gradle-quickstart.sh
matriv: I have added the instructions here: https://github.com/apache/flink/pull/18609/files#diff-b8bb4e796f587464ed5b9e215b5eade045eb10f087df8dfaea260d930aeb684fR415 for how one could use the wrapper.

dmvk: that implies that you already have gradle binaries available locally (so you can use them to generate the wrapper), which for most users won't hold true

matriv: True, otherwise we need to provide the wrapper already in the project generated by the script.

matriv: Or commit the wrapper in the sample project, but I think that's not nice....

matriv: I think it doesn't harm to have the `gradle-wrapper.properties` there, and give an easy possibility to the user to download and use the wrapper. @dmvk what do you think?

dmvk: IMO it does neither harm or help. Personally I'd simply commit the wrapper in the project, all other projects do the same and it doesn't really add any overhead.

matriv:
@dmvk Would you mind checking again, and if you don't object merge it?
I have updated the settings to take into account the flink version regarding the scala/non-scala dependencies,
and the new name of the main Class for `>= 1.15-SNAPSHOT`

q/gradle-quickstart.sh
```diff
@@ -100,53 +103,49 @@ rootProject.name = '${projectName}'
...
+description = """Flink Quickstart Job"""
 
 sourceCompatibility = javaVersion
 targetCompatibility = javaVersion
 tasks.withType(JavaCompile) {
-	options.encoding = 'UTF-8'
+    options.encoding = 'UTF-8'
 }
 
 applicationDefaultJvmArgs = ["-Dlog4j.configuration=log4j.properties"]
```
niklassemmler:
```suggestion
applicationDefaultJvmArgs = ["-Dlog4j.configuration=log4j2.properties"]
```

This appears to be out of date. [See here](https://github.com/apache/flink/tree/master/flink-quickstart/flink-quickstart-java/src/main/resources/archetype-resources/src/main/resources)

matriv: Fixed, we needed also `-Dlog4j.configurationFile` instead of `-Dlog4j.configuration` to make it work with log4j2

q/gradle-quickstart.sh
```diff
@@ -169,17 +168,23 @@ dependencies {
...
     // shadow jar and are provided in the lib folder of Flink
     // --------------------------------------------------------------
-    compile "org.apache.flink:flink-java:\${flinkVersion}"
-    compile "org.apache.flink:flink-streaming-java_\${scalaBinaryVersion}:\${flinkVersion}"
+    if (flinkVersionNew) {
+        implementation "org.apache.flink:flink-streaming-java:\${flinkVersion}"
+        implementation "org.apache.flink:flink-clients:\${flinkVersion}"
+    } else {
+        implementation "org.apache.flink:flink-streaming-java_\${scalaBinaryVersion}:\${flinkVersion}"
+        implementation "org.apache.flink:flink-clients_\${scalaBinaryVersion}:${flinkVersion}"
```
dmvk:
```suggestion
        implementation "org.apache.flink:flink-clients_\${scalaBinaryVersion}:\${flinkVersion}"
```

matriv: Thx for catching that, but somehow, when testing the script it worked without the escape.
Anyways, fixed it, Thx!

dmvk (approved):

PR #507, opened 2022-02-16
[FLINK-26200] Redirect statefun directly to docs

sjwiesman:


igalshilman (approved):
Thanks @sjwiesman, there is no need to keep two different out of sync landing pages anymore.

PR #513, opened 2022-03-14
Restore verification for Google Search Console

MartijnVisser:
Verification was previously done via Google Analytics, but since we have removed that the verification was also removed. This PR restores that verification by including a simple HTML file which Google periodically checks.


knaufk (approved):
LGTM

PR #527, opened 2022-04-12
[FLINK-26446] Update Feature Radar in Apache Flink Roadmap for the 1.15 release

joemoe:


MartijnVisser (changes requested):
Thanks for the update @joemoe 
I think we should make some more changes:
- Kafka, File [via Unified Sink API] should go to Production Ready & Evolving. We should also add Pulsar there. 
- We're missing Kinesis and Firehose as a connector, should probably be in Beta. 
- NiFi Source should be removed from list all together, since we've removed it (it's under Beta)
- DataStream (batch) should move to Production Ready & Evolving
- I would move JDBC Sink to Approaching End-of-Life (because we need to replace it with a JDBC Sink via Unified Sink API
- Same for RabbitMQ source, will also be replaced with one via Unified Source API
- Same for PubSub Source & Sink
- Same for HBase SQL Source & Sink
- Same for Cassandra Sink
- Scala 2.11 can be removed from the diagram, since it has been removed already

joemoe:
@MartijnVisser I addressed your comments, can you please have another look.

MartijnVisser (changes requested):
Sorry @joemoe I now notice that we have Kinesis twice (once under Beta, once under Prd ready & evolving). I think we rename and move Kinesis Source & Sink to Legacy Kinesis Source & Sink and to Deprecated. Kinesis from Beta should be Kinesis, Firehose [via Unified Sink API] and Kinesis [via Unified Source API]

There's also still Elastic Search Sink under Stable, which should be removed. We should add Elasticsearch to the "Kafka, File, Pulsar [via Unified Sink API]" part. Perhaps we should actually prefix the other connectors from EOL (JDBC, RabbitMQ, PubSub, HBase, Cassandra) to 'Legacy' to mark them clearly as the "old" connectors.

The text changes are 👍

joemoe:
@MartijnVisser addressed your comments, can you please have another look? 👁️

MartijnVisser (approved):
@joemoe Thanks Joe! LGTM

mzuehlke:
Should the update date been updated to a more recent date ?
```
**Last Update:** 2021-09-16
```

MartijnVisser:
@mzuehlke Yes, it should have been. Let me fix that :)

PR #543, opened 2022-06-01
[FLINK-27722] Add Slack community information and invite link

MartijnVisser:
@xintongsong If you have a Chinese translation, I could also directly add the Chinese translation to the page.

### DIFF
--- a/community.md
+++ b/community.md
@@ -8,7 +8,7 @@ title: "Community & Project Info"
 
 ## How do I get help from Apache Flink?
 
-There are many ways to get help from the Apache Flink community. The [mailing lists](#mailing-lists) are the primary place where all Flink committers are present. For user support and questions use the *user mailing list*. Some committers are also monitoring [Stack Overflow](http://stackoverflow.com/questions/tagged/apache-flink). Please remember to tag your questions with the *[apache-flink](http://stackoverflow.com/questions/tagged/apache-flink)* tag. Bugs and feature requests can either be discussed on the *dev mailing list* or on [Jira]({{ site.jira }}). Those interested in contributing to Flink should check out the [contribution guide](/contributing/how-to-contribute.html).
+There are many ways to get help from the Apache Flink community. The [mailing lists](#mailing-lists) are the primary place where all Flink committers are present. For user support and questions use the *user mailing list*. You can also join the community on [Slack](#slack). Some committers are also monitoring [Stack Overflow](http://stackoverflow.com/questions/tagged/apache-flink). Please remember to tag your questions with the *[apache-flink](http://stackoverflow.com/questions/tagged/apache-flink)* tag. Bugs and feature requests can either be discussed on the *dev mailing list* or on [Jira]({{ site.jira }}). Those interested in contributing to Flink should check out the [contribution guide](/contributing/how-to-contribute.html).
 
 
 ## Mailing Lists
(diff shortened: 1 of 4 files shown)


community.md
```diff
@@ -144,6 +144,18 @@ If you send us an email with a code snippet, make sure that:
 3. you keep formatting when pasting code in order to keep the code readable
 4. there are enough import statements to avoid ambiguities
 
+## Slack
+
+You can join the [Apache Flink community on Slack.](https://join.slack.com/t/apache-flink/shared_invite/zt-19uavkkkw-qwMimi9gawIv9Fz9xFuqrw)
```
rmetzger, line 149: I wonder if it makes sense to use the ASF URL shortener: https://s.apache.org/
It allows updating the underlying link (however only for the owner of the id) .. but it would still be easier than merging a change to flink-web after 100 invites.

community.md
```diff
@@ -144,6 +144,18 @@ If you send us an email with a code snippet, make sure that:
...
+After creating an account in Slack, don't forget to introduce yourself in #introductions.
+In case the invite link no longer works, please reach out to the [Dev mailing list](#mailing-Lists). 
+
+There are a couple of community rules: 
+* **Be respectful** - This is the most important rule!
+* All important decisions and conclusions **must be reflected back to the mailing lists.** 
+  "If it didn’t happen on a mailing list, it didn’t happen." - The Apache Mottos
```
rmetzger:
```suggestion
  "If it didn’t happen on a mailing list, it didn’t happen." - [The Apache Mottos](http://theapacheway.com/on-list/)
```

community.md
```diff
@@ -144,6 +144,18 @@ If you send us an email with a code snippet, make sure that:
 3. you keep formatting when pasting code in order to keep the code readable
 4. there are enough import statements to avoid ambiguities
 
+## Slack
+
+You can join the [Apache Flink community on Slack.](https://join.slack.com/t/apache-flink/shared_invite/zt-19uavkkkw-qwMimi9gawIv9Fz9xFuqrw)
+After creating an account in Slack, don't forget to introduce yourself in #introductions.
+In case the invite link no longer works, please reach out to the [Dev mailing list](#mailing-Lists).
```
rmetzger:
```suggestion
Due to Slack limitations the invite link expires after 100 invites. If it is expired, please reach out to the [Dev mailing list](#mailing-Lists). 
```
Maybe it makes sense to make this more explicit, so that people understand why this happens?

rmetzger (approved):
Thanks a lot for adding Slack to the website!

community.md
xintongsong: +1

community.md
```diff
@@ -144,6 +144,18 @@ If you send us an email with a code snippet, make sure that:
...
 4. there are enough import statements to avoid ambiguities
 
+## Slack
+
+You can join the [Apache Flink community on Slack.](https://join.slack.com/t/apache-flink/shared_invite/zt-19uavkkkw-qwMimi9gawIv9Fz9xFuqrw)
+After creating an account in Slack, don't forget to introduce yourself in #introductions.
+In case the invite link no longer works, please reach out to the [Dev mailing list](#mailing-Lists). 
+
+There are a couple of community rules: 
+* **Be respectful** - This is the most important rule!
```
xintongsong: We need an empty line here, otherwise the rules are shown as a paragraph rather than a list.
```suggestion
There are a couple of community rules: 

* **Be respectful** - This is the most important rule!
```

community.md
xintongsong: From my side, I'd be more concerned if the invite link can only be updated by one person, compared to having to change the website once a while.

rmetzger: If that person fails to update the link, we can still update the website.
And that person should be one of the three folks involved in this PR ;)

xintongsong, line 149: The invite link is not the only way to join the workspace. Every existing member is free to invite new members, without needing approval from admins. We may want to mention and even suggest that. Hopefully that reduces the frequency we have to update the invite link.

xintongsong:
> @xintongsong If you have a Chinese translation, I could also directly add the Chinese translation to the page.

Sure. Let's address the comments for the English version first, and I'll add a translation before merging.

community.md
MartijnVisser: I agree with @xintongsong that I don't think that having one person who can update the shortened URL is a good idea. I rather just update the Slack invite link (for now) in the project website, since any committer can fix that.

rmetzger: Allright 👍🏻

MartijnVisser: Good point, I've added a comment for that.

xintongsong (approved):
Thanks @MartijnVisser. LGTM.

I've made a few minor changes (see the fixup commit) and added the Chinese translation.

community.md
wuchong: Shall we add a note here for how to reach out to the community that the invitation link hit the limitation?

PR #552, opened 2022-06-22
[hotfix][docs] Eliminate the redundant symbol '/' from the jump link url

yangjf2019:
When the user clicks on the 'Getting started' --> 'with flink' jump in https://flink.apache.org/, there is a redundant '/' symbol in the url, which I have eliminated.

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -115,7 +115,7 @@
             <li class="dropdown">
               <a class="dropdown-toggle" data-toggle="dropdown" href="#">Getting Started<span class="caret"></span></a>
               <ul class="dropdown-menu">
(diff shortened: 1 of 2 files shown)


MartijnVisser (commented):
@yangjf2019 I don't think there's much additional value in this, since it doesn't result in a 404  page.

yangjf2019:
Hi @MartijnVisser ,Thanks for your review. You are right!There is  not much additional value in this.But the dike of a thousand miles is collapsed by an ant hole!I'm just trying to fix this little problem.

MartijnVisser:
I don't see it as a problem. Like I said, all links are currently working. Since work is being done on https://issues.apache.org/jira/browse/FLINK-22922 I think we should just close this PR for now.

PR #563, opened 2022-08-12
[FLINK-28930] Add 'What is Flink Table Store?' link to flink website

SteNicholas:
Similar to statefun and ml projects we should also add a "What is Flink Table Store?" link to the menu pointing to the doc site.

**The brief change log**

- Add 'What is Flink Table Store?' link to flink website.


JingsongLi (approved):
Looks good to me!

PR #577, opened 2022-10-18
[FLINK-29668] Remove Gelly

MartijnVisser:


rmetzger (approved):
I guess the feature radar also needs an update.

roadmap.md
```diff
@@ -176,7 +176,12 @@ trying to make it faster, more predictable, and to remove some confusions and in
 
 There is almost no use case in which Apache Flink is used on its own. It has established itself
 as part of many data related reference architectures. In fact you'll find the squirrel logo covering
-several aspects. The community has added a lot of connectors and formats. With the already mentionend
+several aspects. 
+
+The Flink community has removed Gelly, it's old graph-processing library. There are plans for replacing Gelly based on 
+the new iterations framework that is available via [Flink ML](https://nightlies.apache.org/flink/flink-ml-docs-stable/docs/development/iteration/).
```
zentol: If we write down that there _are plans_ then we better have a place to collect ideas etc. and some people who to signed up for it and a timeline.

zentol: Note that IMO this whole thing about "using iterations framework from Flink ML" is also way too technical for users.

MartijnVisser: > If we write down that there _are plans_ then we better have a place to collect ideas etc. and some people who to signed up for it and a timeline.

I agree. I've pointed to the people on the Dev ML that they should reply on this PR and they should provide input:
- Where ideas can be shared
- Who can be contacted
- What is an expected timeline

If such info can't be provided, that I don't think it makes much sense to include a note on continuation of graph-processing report.

usecases.md
```diff
@@ -67,7 +67,7 @@ Another aspect is a simpler application architecture. A batch analytics pipeline
...
+Flink provides very good support for continuous streaming as well as batch analytics. Specifically, it features an ANSI-compliant SQL interface with unified semantics for batch and streaming queries. SQL queries compute the same result regardless whether they are run on a static data set of recorded events or on a real-time event stream. Rich support for user-defined functions ensures that custom code can be executed in SQL queries. If even more custom logic is required, Flink's DataStream API or DataSet API provide more low-level control.
```
carp84, line 70: Maybe also worth mentioning graph processing here like below?

Moreover, Flink could provide algorithms and building blocks for large-scale and high-performance graph processing, which used to be supported by the Gelly library. However, along with the evolution of Flink (especially the deprecation of the old dataset API), Gelly has been dropped and will be replaced by new implementations based on the new iterations framework via [Flink ML](https://nightlies.apache.org/flink/flink-ml-docs-stable/docs/development/iteration/).

And the Chinese version:

此外，Flink 具备为大规模高性能图分析提供算法和构建支持的能力。Flink中的Gelly模块曾经提供过这种能力，但随着Flink的演进（尤其是dataset API的废弃）Gelly已经被移除，图处理能力将由基于[Flink-ML中的新迭代框架](https://nightlies.apache.org/flink/flink-ml-docs-stable/docs/development/iteration/)的全新实现提供，敬请期待。

carp84 (commented):
Overall LGTM. Added some suggestions, please check the inline comments, thanks. @MartijnVisser

usecases.md
MartijnVisser: > Maybe also worth mentioning graph processing here like below?

But do we really offer good graph processing support? I'm not sure we currently do. I think we can mention that we're planning to improve on our graph processing support, but then at least we should be able to point the user to a FLIP or a Jira ticket.

lindong28: I think the [iteration](https://nightlies.apache.org/flink/flink-ml-docs-stable/docs/development/iteration/) capability supported in Flink ML can not be used to support typical graph processing capabilities (e.g. vertex-centric or gather-sum-apply). And there is currently no plan to support these graph processing capabilities in Flink ML.

So it seems better not to add the paragraph suggested above.

PR #579, opened 2022-11-02
Add Elasticsearch connector 3.0.0

zentol:
Also extends the download page to list connector releases.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -105,6 +105,18 @@ flink_releases:
       sha512_url: "https://downloads.apache.org/flink/flink-1.15.2/flink-1.15.2-src.tgz.sha512"
     release_notes_url: "https://nightlies.apache.org/flink/flink-docs-release-1.15/release-notes/flink-1.15"
 
+connector_releases:
+  - version: "3.0.0"
+    source_release:
+        name: "Apache Flink Elasticsearch Connector 3.0.0"
+        id: "300-elasticsearch-connector-download-source"
+        flink_versions:
+          - "1.16"
+        url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-elasticsearch-3.0.0/flink-connector-elasticsearch-3.0.0-src.tgz"
+        asc_url: "https://downloads.apache.org/flink/flink-connector-elasticsearch-3.0.0/flink-connector-elasticsearch-3.0.0-src.tgz.asc"
+        sha512_url: "https://downloads.apache.org/flink/flink-connector-elasticsearch-3.0.0/flink-connector-elasticsearch-3.0.0-src.tgz.sha512"
+
+
 flink_statefun_releases:
   -
       version_short: "3.2"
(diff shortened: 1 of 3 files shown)


_config.yml
```diff
@@ -611,6 +623,12 @@ release_archive:
         version_long: 2.0.0
         release_date: 2020-04-02
 
+    connectors:
+      - name: "Flink Elasticsearch Connector"
+        connector: "elasticsearch"
+        version: 3.0.0
+        release_date: 2022-11-05
```
dannycranmer: Do we need to include Flink version here? Thinking it might be useful for the case when `3.0.0-1.16` is live and Flink 1.17 is released. We would follow with `3.0.0-1.17`. In this case we have two `3.0.0` versions of the connector and a single release date listed in the docs. We might want to see:

- 3.0.0 for Flink 1.15 and 1.16 released on 2022-11-05
- 3.0.0 for Flink 1.17 released on 2023-01-28

downloads.md
```diff
@@ -107,6 +107,28 @@ Please have a look at the [Release Notes for Flink {{ flink_release.version_shor
 
 {% endfor %}
 
+## Flink connectors
```
dannycranmer, line 110: You missed `downloads.zh.md`

_config.yml
zentol: That would still be one release. We'd just publish an additional set of jars afterwards.

downloads.md
zentol: Left it intentionally in case changes are requested. Will sync it in before the merge.

_config.yml
dannycranmer: In that case we should consider how we announce it. Users might be waiting for certain connectors and will need a means of notification. Maybe a slack message on announce and email to the mailing list will suffice. 

Tangential: In terms of the release itself, I assume we will still require some form of sanity check. We will still need to check the artifacts are good and signing keys are correct. Given the guidance on the [Flink wiki](https://cwiki.apache.org/confluence/display/FLINK/Verifying+a+Flink+Release): 
>  Some of these are strictly required for a release to be ASF compliant.

dannycranmer (approved):

_config.yml
zentol: >  In terms of the release itself, I assume we will still require some form of sanity check. 

That's entirely up to us; there are no requirements from the ASF side on that matter.

MartijnVisser (approved):

_config.yml
zentol: As for announcing it, twitter/slack messages should be good for this. They are well suited for "bursty" announcement, which these will likely be.

Alternatively we could wait with the first big Flink announcement until we have released all other connectors for said version.
So, 1.17 is released (without much fanfare) -> release connectors (without much fanfare) -> announce release :tada: :tada: .
Anyway, this is something better discussed on the ML.

PR #617, opened 2023-03-09
Always pull latest Hugo image on docker build

dannycranmer:
Currently the `./docker-build.sh` script uses the `latest` tag for Hugo docker image. This will use whichever image is installed locally and can cause unexpected changes to the generated site. This change will ensure we are always using the latest image.


hlteoh37 (approved):
LGTM

MartijnVisser (approved):

PR #626, opened 2023-03-27
[FLINK-31627][docs] Updates docker command for website built

XComp:
The user and group ID need to be explicitly set for Linux environments. In Linux, Docker
is usually run as a daemon with root access. Files that are created are, therefore, created
with root being the owner which causes problems trying to move those files into the content*
subfolders.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -20,14 +20,16 @@
 SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
 source "${SCRIPT_DIR}/_utils.sh"
 
-docker pull jakejarvis/hugo-extended:latest
+hugo_docker_image="jakejarvis/hugo-extended:latest"
+docker pull "${hugo_docker_image}"
 
+run_hugo_in_docker="docker run -v $(pwd)/docs:/src -p 1313:1313 ${hugo_docker_image}"
 action=$1
 if [[ $action = "build" ]]
 then
   prepareDocBuild
-  docker run -v $(pwd)/docs:/src -p 1313:1313 jakejarvis/hugo-extended:latest --destination target
+  $run_hugo_in_docker --destination target
   finalizeDocBuild
 else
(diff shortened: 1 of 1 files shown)


XComp:
@MartijnVisser I assume that you're using a MacBook where this issue didn't pop up when migrating to hugo. Am I assuming correctly? Apparently, this isn't an issue in MacOS due to the way how Docker is implemented there.

zentol:
I assumed this to be an issue with my local docker setup.

We didn't have to set users and all that back when we used docker for building the website with jekyll. Why do we need this now?

XComp:
That's actually a good question. I struggled to build the website and came across [this blog post](https://jtreminio.com/blog/running-docker-containers-as-current-host-user/#ok-so-what-actually-works) which sounded reasonable. I didn't check whether we did it like that in the past.

zentol:
This is what we did in the past:
```
docker run --rm --volume="$PWD:/srv/flink-web" --expose=4000 -p 4000:4000 -it ruby:2.6.0 bash -c "cd /srv/flink-web && gem install bundler && ./build.sh $@"
```

Maybe it has something to do with the image itself, and the ruby folks did something special to make this work.

But a quick look in either image didn't reveal anything.

MartijnVisser:
> We didn't have to set users and all that back when we used docker for building the website with jekyll. Why do we need this now?

Because we didn't copy files with Jekyll, but we do now with Hugo (because we manually need to copy over some files, like the htaccess, quickstarts and Google Search Console file)

XComp:
That's a good point: The error only occurs when trying to move the files into the `./content*` directories.

zentol:
> Because we didn't copy files with Jekyll, but we do now with Hugo

Went back to hell and built the website with jekyll, and indeed all created files belong to root.
So this isn't a new problem, we just never ran _really_ into it.
Doubly-so because when git checks out the repo, usually already containing the .html files, they don't belong to root and are just being _modified_ by jekyll.

zentol (approved):

PR #638, opened 2023-04-13
Add JDBC v3.1.0 for Flink 1.16.x and Flink 1.17.x

dannycranmer:


docs/data/release_archive.yml
```diff
@@ -428,6 +428,11 @@ release_archive:
       version: 1.0.0
       release_date: 2022-12-21
       filename: "opensearch"
+    - name: "Flink JDBC Connector"
+      connector: "jdbc"
+      version: 3.0.0
```
elphastori: 3.1.0

elphastori (changes requested):
Checked local build, archive version needs updating. The rest looks good.

docs/data/release_archive.yml
```diff
@@ -428,6 +428,11 @@ release_archive:
...
+      release_date: 2023-04-13
```
libenchao: Do we need to make sure that it's the same with the time we publish to Nexus?

MartijnVisser (approved):
LGTM

elphastori (approved):

docs/data/release_archive.yml
leonardBang: Add a FIXME note is more suggested

leonardBang (approved):

PR #644, opened 2023-04-17
Update the expired slack invitation link

Myasuka:
Since the slack invitation link has been expired, update it.


San-ctuary (approved):

San-ctuary (approved):
the new link is valid

PR #651, opened 2023-05-22
[FLINK-32145] Fix incorrect docker image links of specific versions

Myasuka:
Previous links would point an incorrect locations.


yuchengxin (approved):
LGTM

PR #655, opened 2023-05-25
 Add Pulsar v3.0.1 for Flink 1.16.x

leonardBang:


tisonkun (approved):

docs/data/release_archive.yml
```diff
@@ -490,6 +490,11 @@ release_archive:
       version: 3.0.1
       release_date: 2023-05-09
       filename: "gcp-pubsub"
+    - name: "Flink Pulsar Connector"
```
JingGe: NIT: should we keep the released connectors sorted alphabetically? It will be easier for devs to maintain. WDYT?

wuchong (approved):
LGTM

docs/data/release_archive.yml
leonardBang: It's order by release date, I think it‘s fine to keep current order.

PR #656, opened 2023-05-26
fix zh logo redirect to wrong url

Matrix42:


MartijnVisser (changes requested):
@Matrix42 This won't work, as the HTML files are generated by Hugo

PR #665, opened 2023-07-26
[FLINK-32611] Redirect to Apache Paimon's link instead of legacy flink table store

Myasuka:
Redirect to Apache Paimon's link instead of legacy flink table store.


Myasuka:
@JingsongLi would you please take a look？

JingsongLi:
Can you solve the conflicts?

docs/content.zh/documentation/flink-table-store-0.3.md
```diff
@@ -1,7 +1,7 @@
 ---
-weight: 10
-title: Table Store Master (snapshot)
```
JingsongLi: Maybe we can just delete this file?
We just need to mention Paimon now.

docs/content.zh/what-is-flink-table-store.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: What is Flink Table Store?
+title: What is Paimon(incubating) (formerly Flink Table Store)?
 bookCollapseSection: false
 weight: 5
-bookHref: "https://nightlies.apache.org/flink/flink-table-store-docs-stable/"
+bookHref: "https://paimon.apache.org/docs/master/"
```
JingsongLi: maybe just `https://paimon.apache.org/`?

docs/content.zh/what-is-flink-table-store.md
```diff
@@ -24,6 +24,6 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-# What is the Flink Kubernetes Operator?
+# What is Apache Paimon (formerly Flink Table Store)?
 
-All information on the Flink Table Store can be found on the [Flink Table Store website.](https://nightlies.apache.org/flink/flink-table-store-docs-stable/)
\ No newline at end of file
+The Flink Table Store had joined Apache Incubator as Apache Paimon(incubating). All information on the Apache Paimon(incubating) can be found on the [Paimon website.](https://paimon.apache.org/docs/master/)
```
JingsongLi: Maybe just https://paimon.apache.org/ ?

docs/content/documentation/flink-table-store-0.3.md
```diff
@@ -1,7 +1,7 @@
 ---
-weight: 10
-title: Table Store Master (snapshot)
-bookHref: "https://nightlies.apache.org/flink/flink-table-store-docs-master/"
```
JingsongLi: We can just remove this file.

docs/content/what-is-flink-table-store.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: What is Flink Table Store?
+title: What is Paimon(incubating) (formerly Flink Table Store)?
 bookCollapseSection: false
 weight: 5
-bookHref: "https://nightlies.apache.org/flink/flink-table-store-docs-stable/"
+bookHref: "https://paimon.apache.org/docs/master/"
```
JingsongLi: Maybe just https://paimon.apache.org/ ?

docs/content/what-is-flink-table-store.md
```diff
@@ -24,6 +24,6 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-# What is the Flink Kubernetes Operator?
+# What is Apache Paimon (formerly Flink Table Store)?
 
-All information on the Flink Table Store can be found on the [Flink Table Store website.](https://nightlies.apache.org/flink/flink-table-store-docs-stable/)
\ No newline at end of file
+The Flink Table Store had joined Apache Incubator as Apache Paimon(incubating). All information on the Apache Paimon(incubating) can be found on the [Paimon website.](https://paimon.apache.org/docs/master/)
```
JingsongLi: Maybe just https://paimon.apache.org/ ?

docs/content/documentation/flink-table-store-0.3.md
Myasuka: I prefer to keep this file as flink-table-store was part of history of Flink, and developers/users could still find the valid legacy links if necessary.

docs/content.zh/documentation/flink-table-store-0.3.md
Myasuka: Since flink-table-store-0.3 is the last version when Paimon is part of Flink and cannot be known from Apache Paimon's website. I prefer to keep the link here if someone wants to review the history of Apache Paimon.

docs/content/documentation/flink-table-store-0.3.md
MartijnVisser: Why? With Paimon now becoming its own project, I don't think we should per se link to it from the Flink project website. 

The history of Flink can always be found in Git anyway.

Myasuka: ~~I think we should still host the docs of legacy version of Paimon for users, and it seems we cannot find the legacy docs of older version via Paimon's website. That's why I prefer to keep it.~~

**Update:** I found we can get the link from [Paimon's website](https://paimon.apache.org/docs/0.4/versions/), even it is not easy to find, I think users could still get it. If so, I think we can remove the legacy link and keep the link to the latest Paimon's doc.

MartijnVisser: I understand that, but I don't think we should link to the legacy docs from the main menu. It will just redirect users to outdated content, while they should be checking out the latest docs. 

Paimon could consider linking to the old docs from their project website. Another approach would be that the announcement of Paimon itself would be posted as a Flink blog, and that includes links to the old doc versions there.

Myasuka:
Already updated this PR, please take a review again @JingsongLi @MartijnVisser .

carp84:
@Myasuka I respect the consideration of recording the history, whereas the history is already recorded in the below ways (which explicitly show that Paimon origins from Flink Table Store), besides this PR:

1. The incubation [proposal](https://cwiki.apache.org/confluence/display/INCUBATOR/PaimonProposal) and [discussion](https://lists.apache.org/thread/hr3d7tpw02w6ybrnnlf3hcbhfxotwpvn) of Paimon
2. The [announcement](https://lists.apache.org/thread/pz5f9cvpyk4q9vltd7z088q5368v412t) that Flink Table Store joins Apache incubator as Paimon

And we will send another email to our Flink user/dev mailing lists if one day Paimon graduates as a Top-Level-Project (you have my word).

It's not bad to be nostalgic, but let's move forward to the future (smile)

Myasuka:
@carp84 Thanks for the review, I had already removed the links to the legacy Paimon (flink-table-store) and only left a link to the incubating-paimon doc. From the discussion, I think guys have similar ideas to drop the legacy links, and it seems no obvious opinion against to link to the latest paimon docs.

JingsongLi (approved):
Looks good to me!

PR #667, opened 2023-07-31
[hotfix] Update latest stable version to 1.17.1

patricklucas:


RyanSkraba (approved):
LGTM -- this only seems to appear in the https://flink.apache.org/downloads/ page, but it should be changed so the maven artfact snippets are correct.

davidradl:
This is outdate now - there is a 1.17.2, [patricklucas](https://github.com/patricklucas) do you want to update or close?

patricklucas:
yeah this was from 2.5 years ago... hadn't realized it was never merged.

PR #673, opened 2023-09-06
Fix the search bar on Flink website

dannycranmer:
Currently the search bar loads pages to search.

### DIFF
--- a/docs/assets/search-data.js
+++ b/docs/assets/search-data.js
@@ -36,15 +36,20 @@
 
   {{- $pages := where .Site.Pages "Kind" "in" (slice "page" "section") -}}
   {{- $pages = where $pages "Params.booksearchexclude" "!=" true -}}
-  {{- $pages = where $pages "Content" "not in" (slice nil "") -}}
 
-  {{ range $index, $page := $pages }}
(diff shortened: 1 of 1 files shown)


docs/assets/search-data.js
```diff
@@ -36,15 +36,20 @@
 
   {{- $pages := where .Site.Pages "Kind" "in" (slice "page" "section") -}}
   {{- $pages = where $pages "Params.booksearchexclude" "!=" true -}}
-  {{- $pages = where $pages "Content" "not in" (slice nil "") -}}
```
dannycranmer, line 39: @MartijnVisser I could not work out how to get this filter fixed. I tried a bunch of things, including inverting the logic from `not in` to `in`, and simplifying to `(slice nil)`/`(slice "")`.  Could not get any results. I have refactored to a non-ideal loop with filter using a counter to generate the index. It works..... but it is not nice :D 

Any idea?

foxus:
@MartijnVisser, I spent some time with @dannycranmer attempting to debug the behaviour of the Content filter to no avail. Whilst the filter and counter implementation is verbose, it is performant. LGTM.

PR #676, opened 2023-09-26
[FLINK-33046][FLIP-333] Redesign Apache Flink website

dannycranmer:
Applied styling/structural changes as per the design in https://cwiki.apache.org/confluence/display/FLINK/FLIP-333%3A+Redesign+Apache+Flink+website

See a hosted version of this branch [here](https://website-refresh.d193kg429zpv7e.amplifyapp.com/).

Light and dark mode are supported, to toggle please change your system preferences.


dannycranmer:
I am waiting for new blog/twitter/github icons for the footer.

docs/i18n/en.yaml
```diff
@@ -0,0 +1,163 @@
...
+
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+- id: hero.title
+  translation: Apache Flink
```
rmetzger, line 19: The old website is using the trademark (r) for the first mention of Apache Flink on the website. In my understanding of the trademark usage rules, we should do that on our own website.

docs/i18n/en.yaml
```diff
@@ -0,0 +1,163 @@
...
+  translation: Recent Flink blogs
+
+- id: footer.disclaimer
+  translation: The contents of this website are © 2023 Apache Software Foundation under the terms of the Apache License v2. Apache Flink, Flink, and the Flink logo are either registered trademarks or trademarks of The Apache Software Foundation in the United States and other countries.
+
+- id: footer.powered-by
+  translation: Powered by
+
+- id: footer.flink-package
+  translation: flink-package.org
```
rmetzger:
```suggestion
  translation: flink-packages.org
```

docs/i18n/zh.yaml
```diff
@@ -0,0 +1,163 @@
...
+  translation: 最新博客列表
+
+- id: footer.disclaimer
+  translation: The contents of this website are © 2023 Apache Software Foundation under the terms of the Apache License v2. Apache Flink, Flink, and the Flink logo are either registered trademarks or trademarks of The Apache Software Foundation in the United States and other countries.
+
+- id: footer.powered-by
+  translation: Flink 用户
+
+- id: footer.flink-package
+  translation: flink-package.org
```
rmetzger:
```suggestion
  translation: flink-packages.org
```

rmetzger:
Why is "Powered By" in the footer and not in the "About" menu?

rmetzger:
Is the security page gone? https://flink.apache.org/security/

docs/i18n/en.yaml
dannycranmer: Ack, will copy the old site and put it with the home page title

dannycranmer:
> Why is "Powered By" in the footer and not in the "About" menu?

This is as per the [FLIP design](https://cwiki.apache.org/confluence/display/FLINK/FLIP-333%3A+Redesign+Apache+Flink+website). Would you prefer it in both locations, or just under About?

> Is the security page gone? https://flink.apache.org/security/

It is included, we just missed the link. Since there are multiple "Security" links on the existing site it must have slipped through: https://website-refresh.d193kg429zpv7e.amplifyapp.com/security/ . I will add it under about

rmetzger:
Thank you for addressing my feedback.

> Would you prefer it in both locations, or just under About?

This is a very subjective question, there's no right or wrong here. But I find "the big logo collection" a pretty important argument for proving the maturity of the project, so it should be placed prominently. I would put it into the About menu, below Use Cases, similar to the current web site.
Whether to put it into both locations: Probably its fine to just put it into About.

rmetzger:
I am ready to soft-approve this pull request, but would feel more comfortable if another committer would take a look as well, given that this PR is changing the "face" of Flink pretty significantly.

dannycranmer:
> Thank you for addressing my feedback.
> 
> > Would you prefer it in both locations, or just under About?
> 
> This is a very subjective question, there's no right or wrong here. But I find "the big logo collection" a pretty important argument for proving the maturity of the project, so it should be placed prominently. I would put it into the About menu, below Use Cases, similar to the current web site. Whether to put it into both locations: Probably its fine to just put it into About.

Done. Thanks for the feedback

MartijnVisser (changes requested):
Some comments:

1. Not sure if it's intended, but when hovering over the Flink logo in top left, then "Apache Flink" becomes visible (in dark mode). I think we should always display it, because the sudden appearance after a hover feels wrong? 
image
image

2. The menu isn't on the left side like in the FLIP, but on the right hand side
image

3. When a link has been visited, we're using a different color in the menu. Can we just keep the same color as it is initially? This looks messy
image

dannycranmer:
> Not sure if it's intended, but when hovering over the Flink logo in top left, then "Apache Flink" becomes visible (in dark mode). I think we should always display it, because the sudden appearance after a hover feels wrong?

This is not intentional. I cannot reproduce this, I am also using Chrome on Mac. Any other details you can provide to help me reproduce? Chrome version?

> The menu isn't on the left side like in the FLIP, but on the right hand side

We put it on the left initially and internal feedback was that users thought it was a nav, rather than page index. We moved to right as per this feedback, this also aligns with the previous website. What do you think? I can raise this on the discussion thread if you think it is worth it.

> When a link has been visited, we're using a different color in the menu. Can we just keep the same color as it is initially? This looks messy

Ack, I agree, will do. How about link within a page?

Myasuka:
I found a strange behavior that when I click the nav title on the right, it will hide the head of lines:
image
What I expect:
image

The problematic link https://website-refresh.d193kg429zpv7e.amplifyapp.com/what-is-flink/flink-applications/#state

MartijnVisser:
> This is not intentional. I cannot reproduce this, I am also using Chrome on Mac. Any other details you can provide to help me reproduce? Chrome version?

Open any Staging link in an incognito window. Using `Version 117.0.5938.149 (Official Build) (arm64)` 

> We put it on the left initially and internal feedback was that users thought it was a nav, rather than page index. We moved to right as per this feedback, this also aligns with the previous website. What do you think? I can raise this on the discussion thread if you think it is worth it.

Hmmm. I guess I just miss the menu on the left. I don't like it but won't block it.

> Ack, I agree, will do. How about link within a page?

I think we just shouldn't use the distinguishment at all if you've already visited a link before.

dannycranmer:
> I found a strange behavior that when I click the nav title on the right, it will hide the head of lines: image What I expect: image
> 
> The problematic link https://website-refresh.d193kg429zpv7e.amplifyapp.com/what-is-flink/flink-applications/#state

@Myasuka  Thanks for reporting this! Fixed and deployed to the demo site.

dannycranmer:
@MartijnVisser thanks for reporting this:

> I think we just shouldn't use the distinguishment at all if you've already visited a link before.

Fixed and deployed to the demo site.

dannycranmer:
> Hmmm. I guess I just miss the menu on the left. I don't like it but won't block it.

We can always change it later if we get similar feedback. Thanks for not blocking

dannycranmer:
> Open any Staging link in an incognito window. Using Version 117.0.5938.149 (Official Build) (arm64)

Thanks, have reproduced in incognito! (weird)

dannycranmer:
@MartijnVisser thanks for reporting this:

> Not sure if it's intended, but when hovering over the Flink logo in top left, then "Apache Flink" becomes visible (in dark mode). I think we should always display it, because the sudden appearance after a hover feels wrong?

Fixed and deployed to the demo site.

MartijnVisser (approved):

hlteoh37 (approved):
LGTM, sanity checked the equivalent Chinese sites (https://flink.apache.org/zh/*). Thanks @dannycranmer !

PR #686, opened 2023-10-13
[hotfix][docs] Second attempt to fix broken links ;)

rmoff:
Following on from #684

- Fix aliases that were redirecting incorrectly to /zh
- Add index.html aliases for the pages too


dannycranmer:
LGTM, verified locally

dannycranmer (approved):

PR #689, opened 2023-10-18
[hotfix][docs] Change how-to-contribute empty link to overview page

swap-10:
The how-to-contribute link on the [what is flink](https://flink.apache.org/what-is-flink/community/#how-do-i-get-help-from-apache-flink) page currently links to /how-to-contribute/
This is an empty page.
This commit changes the link to /how-to-contribute/overview/ which seems to be the intended helpful page.

I considered using an alias instead if possible, but the /what-is-flink URL also currently opens an empty page (no content), except that it's not linked to from anywhere so user experience is not impacted. So I assume this was a conscious design decision and just changed the link in question from
/how-to-contribute/
to
/how-to-contribute/overview/

I have built it locally with Hugo and tested that it works.
Thanks!

### DIFF
--- a/docs/content/what-is-flink/community.md
+++ b/docs/content/what-is-flink/community.md
@@ -122,7 +122,7 @@ under the License.
 
 ## How do I get help from Apache Flink?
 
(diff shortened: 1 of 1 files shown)


RyanSkraba (commented):
Hey -- thanks for the contribution.  I've verified that this doesn't lead to the correct link in the current website, and that your change fixes it.

Can you make the parallel change in the chinese section of the documentation (`content-zh` ) ? https://github.com/apache/flink-web/blob/asf-site/docs/content.zh/what-is-flink/community.md

I'm tempted to say that you've noticed a bug with the empty `_index.md` files (in https://flink.apache.org/how-to-contribute/ for example) which should be rediirected to the first page in their section -- how do you feel about raising a JIRA and fixing these as well?

dannycranmer:
Since there is no reply here, I will merge this and fix the CN myself

dannycranmer:
Done https://github.com/apache/flink-web/commit/eb6412b02e85c9dc7291e5759bb2fb02a259e774

PR #693, opened 2023-10-31
Adding AWS Connectors v4.2.0

dannycranmer:


MartijnVisser (approved):

vahmed-hamdy (approved):

JingGe (approved):
Thanks! LGTM

docs/data/release_archive.yml
```diff
@@ -522,6 +522,11 @@ release_archive:
       version: 3.0.1
       release_date: 2023-10-30
       filename: "kafka"
+    - name: "Flink AWS Connectors"
+      connector: "aws"
```
leonardBang, line 526: Should we list all dedicated connector names instead of a special module name here?

Just a minor comment, not a blocker.

leonardBang (approved):

docs/data/release_archive.yml
dannycranmer: Hey. I like this idea. This bubbles to here: https://flink.apache.org/downloads/#apache-flink-connectors-1 . Was you recommending one line per connector, or to change the "Flink AWS Connectors" to "Flink Amazon Kinesis, DynamoDB, Firehose Connector"? We also have the formats so it might get a bit long

leonardBang: How about simply using YAML sequence like:
```
 - name: "Flink AWS Connectors"
   connector: ["kinesis", "dynamodb","firehose"]
```

PR #701, opened 2023-11-28
Add flink-shaded 18.0 release

snuyanzin:


docs/data/additional_components.yml
```diff
@@ -21,11 +21,11 @@ flink-connector-parent:
...
-  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-shaded-17.0/flink-shaded-17.0-src.tgz"
-  source_release_asc_url: "https://downloads.apache.org/flink/flink-shaded-17.0/flink-shaded-17.0-src.tgz.asc"
-  source_release_sha512_url: "https://downloads.apache.org/flink/flink-shaded-17.0/flink-shaded-17.0-src.tgz.sha512"
+flink-shaded-18.0:
+  name: "Apache Flink-shaded 18.0 Source Release"
+  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-shaded-18.0/flink-shaded-18.0-src.tgz"
+  source_release_asc_url: "https://downloads.apache.org/flink/flink-shaded-18.0/flink-shaded-18.0-src.tgz.asc"
+  source_release_sha512_url: "https://downloads.apache.org/flink/flink-shaded-18.0/flink-shaded-18.0-src.tgz.sha512"
 
 flink-shaded-16.2:
```
1996fanrui: Keep `18.0` and `16.2` isn't a good idea here.

Should we just keep the 17.0 and 18.0(The latest 2 versions)? Or keep all versions here?

1996fanrui (changes requested):

docs/data/additional_components.yml
snuyanzin: That's actually a good question
the only thing making me stopping before removal that is that it was released just a couple of weeks ago...
However I think you are right and  probably we should not stick to it

1996fanrui: I see the `pre-bundled-hadoop-` has 4 versions in this yml, so should we keep 16, 17 and 18?

TanYuxin-tyx: +1, maybe it is a good choice to keep 16.2, 17.0, and 18.0.

snuyanzin: thanks for the feedback, I added 17.0 back to the site

TanYuxin-tyx (approved):
Thanks for the update, LGTM

1996fanrui (approved):
Thanks @snuyanzin for the update and thanks @TanYuxin-tyx for the feedback!

LGTM

MartijnVisser (approved):

PR #702, opened 2023-11-29
Make Flink-specific security page more prominent

raboof:
As this will be relevant to Flink users.

This page then links to the generic information at https://www.apache.org/security


dannycranmer (approved):
Thanks for the contribution @raboof

PR #707, opened 2024-01-05
Add JDBC 3.1.2 connector

MartijnVisser:


docs/data/release_archive.yml
```diff
@@ -545,6 +545,11 @@ release_archive:
       version: 4.1.0
       release_date: 2023-12-28
       filename: "pulsar"
+    - name: "Flink JDBC Connector"
+      connector: "jdbc"
+      version: 3.1.2
+      release_date: 2024-05-01
```
Jiabao-Sun: Does the release date need to be adjusted?

ruanhang1993 (approved):

docs/data/release_archive.yml
leonardBang: +1

docs/data/flink_connectors.yml
```diff
@@ -44,11 +44,11 @@ gcp-pubsub:
...
-  name: "Apache Flink JDBC Connector 3.1.1"
-  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-jdbc-3.1.1/flink-connector-jdbc-3.1.1-src.tgz"
-  source_release_asc_url: "https://downloads.apache.org/flink/flink-connector-jdbc-3.1.1/flink-connector-jdbc-3.1.1-src.tgz.asc"
-  source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-jdbc-3.1.1/flink-connector-jdbc-3.1.1-src.tgz.sha512"
-  compatibility: ["1.16.x", "1.17.x"]
+  name: "Apache Flink JDBC Connector 3.1.2"
+  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-jdbc-3.1.2/flink-connector-jdbc-3.1.2-src.tgz"
+  source_release_asc_url: "https://downloads.apache.org/flink/flink-connector-jdbc-3.1.2/flink-connector-jdbc-3.1.2-src.tgz.asc"
+  source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-jdbc-3.1.2/flink-connector-jdbc-3.1.2-src.tgz.sha512"
+  compatibility: ["1.16.x", "1.17.x", "1.18.x"]
```
leonardBang, line 58: > The flink.version  that's set in the root pom.xml should be set to the lowest Flink version that's supported. You can't use highest because there are no guarantees that something that's working in e.g. Flink 1.18 is working in Flink 1.17.

https://cwiki.apache.org/confluence/display/FLINK/Externalized+Connector+development said above, but here we're actually https://github.com/apache/flink-connector-jdbc/blob/v3.1.2-rc2/pom.xml#L50 , how we ensure we can support Flink 1.16.x?

As the flink version 1.17 has been used in `Apache Flink JDBC Connector 3.1.1`, I agree this is a history issue and not sure it's a blocker or not.

snuyanzin: good catch
it seems same history issue appears for artifacts at 
https://repository.apache.org/content/repositories/orgapacheflink-1704/org/apache/flink/flink-connector-jdbc/
where there are artifacts for 1.16...

I wonder if there is a way to remove it from there without generation a new RC?
Otherwise we probably need another RC

leonardBang: I think we can simply support `compatibility: ["1.17.x", "1.18.x"]` which means we only need to build a new rc2 without change any code, WDYT?

leonardBang: > I wonder if there is a way to remove it from there without generation a new RC?

Maybe the way does not exists, 1.16 related artifacts is one part of the RC1 and maven staging repo,  creating a new RC is  more safe and recommended.

snuyanzin: @leonardBang sorry for the confusion
I noticed that despite there is 1.17 in pom, there is nightly/weekly tests for 1.16 as well [1] so we can confirm that tests for Flink 1.16 are passing.
Based on history it came from FLINK-31699. I assumed it could be as a result of failing archunit tests for 1.16. @MartijnVisser please correct me if I'm wrong

[1] https://github.com/apache/flink-connector-jdbc/actions/runs/7723924144

MartijnVisser: I think it's inadvertently bumped Flink version: since the connector still works with Flink 1.16 (as proven in the nightly builds), we should have kept this `flink.version` set to `1.16.0`. Only when we have to drop support for a lower Flink version, we should bump this.

snuyanzin: I added a PR to fix it in pom and ci https://github.com/apache/flink-connector-jdbc/pull/99

now both `push pr` and `weekly` ci's are running against Flink 1.16 for branch v3.1

leonardBang: please also backport to main branch

leonardBang (approved):
Thanks @snuyanzin for the release work, LGTM

docs/data/flink_connectors.yml
snuyanzin: It is the issue for v3.1  only
for main it was fixed within FLINK-33378 and seems was not backported to v3.1...

leonardBang: Actually I mean  here[1] should be `1.16.0`  instead of `1.16.2`, `1.16.0` is the minimum supported flink version.
[1] https://github.com/apache/flink-connector-jdbc/commit/79e7509256753765813d2e2c970506414de3302a#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R50

leonardBang: You did right fix in v3.1 branch ^_^
https://github.com/apache/flink-connector-jdbc/pull/99/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R50

snuyanzin: ok, got it, thanks
created a PR for that https://github.com/apache/flink-connector-jdbc/pull/102

snuyanzin:
rebased to resolve conflicts

RyanSkraba (approved):
Is checking in the `/content` directory deliberate here?  It will have to be rebuilt regardless after this commit, of course, so no harm done, but for consistency it _could_ be removed here.

I checked the links, rebuilt the website and checked the results.  LGTM, thanks!

***Edit***: My local check used a different `jakejarvis/hugo-extended:latest` than the one used here, it probably does *not* need to be rebuilt after this change.

snuyanzin:
content  changes is a result of website rebuild

snuyanzin (approved):

PR #711, opened 2024-01-12
[hotfix] Use valid url for flink-shaded

snuyanzin:
The problem with website that any flink-shaded link from archive section 
https://flink.apache.org/downloads/#apache-flink-shaded
leads to `Not Found`
The PR fixes this


XComp (approved):
LGTM

PR #732, opened 2024-04-09
[FLINK-34953][ci] Add github ci for flink-web to auto commit build files

GOODBOY008:
Changes:

- Add github ci for flink-web to auto commit build files

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,66 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Flink Web CI"
+on:
+  pull_request:
+  push:
+    branches:
+      - asf-site
+  workflow_dispatch:
+
+jobs:
+  build-documentation:
+    if: github.repository == 'apache/flink-web'
+    runs-on: ubuntu-latest
+    permissions:
+      # Give the default GITHUB_TOKEN write permission to commit and push the changed files back to the repository.
+      contents: write
+    steps:
+        - name: Checkout repository
+          uses: actions/checkout@v4
+          with:
+            submodules: true
+            fetch-depth: 0
+
+        - name: Setup Hugo
+          uses: peaceiris/actions-hugo@v3
+          with:
+            hugo-version: '0.119.0'
+            extended: true
+
+        - name: Build website
+          run: |
+            # Remove old content folder and create new one
+            rm -r -f content && mkdir content
+
+            # Build the website
+            hugo --source docs --destination target
+
+            # Move newly generated static HTML to the content serving folder
+            mv docs/target/* content
+
+            # Copy quickstarts, rewrite rules and Google Search Console identifier
+            cp -r _include/. content
+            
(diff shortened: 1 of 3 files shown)


GOODBOY008:
The Action result: https://github.com/GOODBOY008/flink-web/commits/asf-site/

DEAY~2J7}9_@ZN)225X5FAV

.github/workflows/docs.yml
```diff
@@ -0,0 +1,68 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Flink Web CI"
+on:
+  pull_request:
+    branches:
+      - asf-site
```
MartijnVisser: I don't think this should work on PRs, but only on pushes? Else you would commit code during a PR creation?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,68 @@
...
+
+            # Copy quickstarts, rewrite rules and Google Search Console identifier
+            cp -r _include/. content
+            
+            # Get the current commit author
+            echo "author=$(git log -1 --pretty=\"%an <%ae>\")" >> $GITHUB_OUTPUT
+
+        - name: Commit and push website build
+          if: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
+          uses: stefanzweifel/git-auto-commit-action@v5
```
MartijnVisser: Per https://infra.apache.org/github-actions-policy.html we must review and pin this

.github/workflows/docs.yml
GOODBOY008: I want to enable `website build check` for pr to avoid doc error without auto commit.

MartijnVisser: But in this setup, you will always build and commit the docs as well.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,68 @@
...
+            mv docs/target/* content
+
+            # Copy quickstarts, rewrite rules and Google Search Console identifier
+            cp -r _include/. content
+            
+            # Get the current commit author
+            echo "author=$(git log -1 --pretty=\"%an <%ae>\")" >> $GITHUB_OUTPUT
+
+        - name: Commit and push website build
+          if: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
```
GOODBOY008, line 62: @MartijnVisser With `push` event for pr merge into branch and `workflow_dispatch` for manual trigger to rebuild website.

GOODBOY008:
@MartijnVisser @XComp PTAL

GOODBOY008:
@MartijnVisser Can you give me some suggestions?

GOODBOY008:
@MartijnVisser @XComp  
This change aims to reduce the workload for committers by automating the build and commit process for the website's generated content. Could you kindly review this PR and let me know if you have any questions or concerns? I'm happy to discuss further if needed.  

The PR Funtion review：
Link: https://github.com/GOODBOY008/flink-web/commits/asf-site/
WXWorkCapture_17317273824345

.github/workflows/docs.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  build-documentation:
+    if: github.repository == 'apache/flink-web'
+    runs-on: ubuntu-latest
+    permissions:
+      # Give the default GITHUB_TOKEN write permission to commit and push the changed files back to the repository.
+      contents: write
+    steps:
+        - name: Checkout repository
```
davidradl, line 32: I like this idea. I wonder about. 
- we should not rebuild the website if the change to the website is not a source file 
- so we are clean - maybe we should also reject PR changes that contain any changes to the generated part of the repo.

MartijnVisser:
I'm not personally convinced this is worth the effort, but I don't want that to be the only thing holding it up.

If there's broader appetite for it, the thing to solve first is that our source and our generated output live on the same branch. That's what makes this workflow commit to the branch that triggered it, and it's why generated HTML can end up in PR diffs. NiFi and DataFusion keep those separate; we'd want to do the same before auto-committing is safe. That's a dev@ discussion rather than something to solve in this PR.

Separately: `peaceiris/actions-hugo@v3` needs a full SHA pin per https://infra.apache.org/github-actions-policy.html.

PR #742, opened 2024-05-16
[FLINK-35331] Show binary files to download as binary links and not as source links

hecspc:
Hadoop links have only links to download the binary files. Changed the template to support only binary files so no need to add the source files. This way we can add to the download page only the binary links to download.


MartijnVisser:
> Changed the template to support only binary files so no need to add the source files.

That's not in line with ASF requirements, closing this PR

PR #753, opened 2024-08-15
Add release announcement for CDC 3.2.0

yuxiqian:
This PR adds release announcement post, and updates metadata info for the coming CDC 3.2.0 release.


docs/content/posts/2024-08-15-release-cdc-3.2.0.md
```diff
@@ -0,0 +1,121 @@
...
+Now it is possible to deploy a YAML pipeline job to Kubernetes cluster with CDC CLI. 
+See [Deployment Docs](https://nightlies.apache.org/flink/flink-cdc-docs-release-3.2/docs/deployment/kubernetes/) for more details.
+
+## Customizable Schema Evolution
+
+More customizable schema evolution modes have been added.
+Now, apart from existing "Evolve", "Ignore", and "Exception" mode, a pipeline job could be configured as:
+
+* "TryEvolve": Always do schema evolution attempts, but evolution failures will be tolerated.
+* "Lenient": Similar to "TryEvolve" mode, but provides more lenient schema conversion after failover.
```
leonardBang: Could you give some example to explain the lenient behavior? Eg. what will happen when ``drop column`,`add column` event meets  lenient behavior.

leonardBang: And the default schema evolution behavior should be reminded from my side.

yuxiqian:
Thanks for @leonardBang's review, addressed comments above.

One minor concern: should Flink CDC 3.2.0 be declared as compatible with Flink 1.20? For now it is not possible to restore from checkpoint when using 1.20.

leonardBang:
> Thanks for @leonardBang's review, addressed comments above.
> 
> One minor concern: should Flink CDC 3.2.0 be declared as compatible with Flink 1.20? For now it is not possible to restore from checkpoint when using 1.20.

@yuxiqian As we didn't have tests to cover flink 1.20 well, I think we should not declare this compatibility.

PR #763, opened 2024-11-14
Upgrade Kafka connector to 3.4.0

AHeise:


lvyanquan (approved):
LGTM.

leonardBang (approved):
+1

docs/data/flink_connectors.yml
```diff
@@ -58,11 +58,11 @@ jdbc:
   compatibility: ["1.18.x", "1.19.x"]
 
 kafka:
-  name: "Apache Flink Kafka Connector 3.3.0"
-  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-kafka-3.3.0/flink-connector-kafka-3.3.0-src.tgz"
-  source_release_asc_url: "https://downloads.apache.org/flink/flink-connector-kafka-3.3.0/flink-connector-kafka-3.3.0-src.tgz.asc"
-  source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-kafka-3.3.0/flink-connector-kafka-3.3.0-src.tgz.sha512"
-  compatibility: ["1.19.x", "1.20.x"]
```
MartijnVisser: I believe we still need to leave this in the website, because else there's no connector for Flink 1.19 available, yet that's still supported in the community. So I think in the end we need both 3.3.0 and 3.4.0 to be listed here.

MartijnVisser (approved):
I've left one comment but that's a non-blocked fix

dannycranmer (approved):
LGTM, agree with @MartijnVissers comment

docs/data/flink_connectors.yml
AHeise: Fixed but I wonder if we even need to leave in 3.2.0 for 1.18 support?

PR #782, opened 2025-03-24
[hotfix] Fix the broken TOC

Zakelly:
After flink 2.0 release, the TOC seems broken.
image

This seems to be raised by a link in heading. I removed that and test in my local:

It looks fine. I ran ```./build.sh build``` and commit the changed files.


Zakelly:
The first commit is the core change.

reswqa (approved):
Thanks for the fix.

PR #787, opened 2025-04-08
Upgrade Kafka connector to 4.0.0 for Flink 2.0

AHeise:


docs/data/flink_connectors.yml
```diff
@@ -78,6 +78,13 @@ kafka34:
   source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-kafka-3.4.0/flink-connector-kafka-3.4.0-src.tgz.sha512"
   compatibility: ["1.20.x"]
 
+kafka40:
```
leonardBang, line 95: Should we remove `kafka33` in this PR to keep only two versions?

leonardBang (approved):

docs/data/flink_connectors.yml
leonardBang: minor: `kafka33` , `kafka34` and `kafka40` is a new name style IIRC,  I like it too. Could you broadcast it to all connectors and document in connector release wiki?

snuyanzin (approved):

docs/data/flink_connectors.yml
AHeise: > Should we remove `kafka33` in this PR to keep only two versions?

We need to keep it around until we drop Flink 1.19 support. I'm actually not sure if we already have.

PR #826, opened 2026-02-04
Add Flink Connector Parent 2.0.0

ferenc-csaky:
Release date to be added when vote passes and we deploy the release.


gaborgsomogyi:
Please rebase.

gaborgsomogyi (approved):
The changes looks good

docs/data/additional_components.yml
```diff
@@ -16,10 +16,10 @@
 # under the License
 
 flink-connector-parent:
-  name: "Apache Flink-connector-parent 1.1.0 Source release"
-  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-parent-1.1.0/flink-connector-parent-1.1.0-src.tgz"
-  source_release_asc_url: "https://downloads.apache.org/flink/flink-connector-parent-1.1.0/flink-connector-parent-1.1.0-src.tgz.asc"
-  source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-parent-1.1.0/flink-connector-parent-1.1.0-src.tgz.sha512"
+  name: "Apache Flink-connector-parent 2.0.0 Source release"
```
davidradl: I see some connectors still use 1.0 some 1.1 and we hope for future Flink 2 connectors to use the v2. Should be include all the versions here? If we just have the latest - I am curious why.

ferenc-csaky: That's actually a good point. Until the 1.x major version is still in use, it should have been kept. Updated.

leonardBang (approved):

docs/data/additional_components.yml
```diff
@@ -15,7 +15,13 @@
...
 # under the License
 
-flink-connector-parent:
+flink-connector-parent-2:
+  name: "Apache Flink-connector-parent 2.0.0"
+  source_release_url: "https://www.apache.org/dyn/closer.lua/flink/flink-connector-parent-2.0.0/flink-connector-parent-2.0.0-src.tgz"
+  source_release_asc_url: "https://downloads.apache.org/flink/flink-connector-parent-2.0.0/flink-connector-parent-2.0.0-src.tgz.asc"
+  source_release_sha512_url: "https://downloads.apache.org/flink/flink-connector-parent-2.0.0/flink-connector-parent-2.0.0-src.tgz.sha512"
+
+flink-connector-parent-1:
```
davidradl, line 24: I see the kafka connector still refers to [1.0](https://github.com/apache/flink-connector-kafka/blob/cdbd635171fc4322ba7182eb93e920472d6d9d91/pom.xml#L24) . Do we need to include 1.0 here as well?

ferenc-csaky: I don't think so, 1.0.0 is not supported with 1.1.0 released, so I do not see the point, any connector should update to 1.1.0 at least already.

davidradl: OK I have raised [https://issues.apache.org/jira/browse/FLINK-39070](https://issues.apache.org/jira/browse/FLINK-39070) to track this

PR #827, opened 2026-02-04
Reorder connector releases

ferenc-csaky:
## Key Changes
- Inverted the connector release list to order it by date DESC to match everything else on the downloads page
- Some entries under "Additional Components" on the download page had the phrase "Source Release" duplicated, cause it was explicitly added to the name in the YAML, removed that
- Some title changes to uppercase "Connectors" to match everything else


mbalassi:
Thanks, looks good to me.

PR #837, opened 2026-03-31
[FLINK-39353] Rebuild website with docker which uses fixed hugo version

leonardBang:
Rebuild website with docker which uses fixed hugo version  0.124.1

Checked in my local env as following:

image


lvyanquan (approved):
+1, thank you for the quick fix. I also manually verified it by starting the service with Docker.

1996fanrui:
Thanks @leonardBang for the fix!

PR #838, opened 2026-03-31
[FLINK-39353][build] Check hugo version in local bash scripts

leonardBang:
Check hugo version in local bash scripts to ensure align with docker bash scripts

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,10 +17,28 @@
 # limitations under the License.
 ################################################################################
 
+REQUIRED_HUGO_VERSION="0.124.1"
+
 if ! command -v hugo &> /dev/null
 then
 	echo "Hugo must be installed to run the docs locally"
 	echo "Please see docs/README.md for more details"
+	exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -17,13 +17,31 @@
 # limitations under the License.
 ################################################################################
 
+REQUIRED_HUGO_VERSION="0.124.1"
```
snuyanzin, line 20: I wonder if we can consider this as an opportunity to bump the version
WDYT?

lvyanquan (approved):
LGTM.

build.sh
leonardBang: I also thought bumping hugo version int this PR, but considered the original un-readable dark mode was led by higher hugo version, thus I only want to add a check in this PR and then bump version in another independent task.

snuyanzin: makes sense

snuyanzin (approved):

1996fanrui (approved):
Thanks for the improvement, lgtm.

PR #854, opened 2026-06-01
Rewrite security page with precise trust model and operator coverage

gaborgsomogyi:
The previous page had three problems: a stale CVE table that could not realistically be kept up-to-date, a vague security model that generated false vulnerability reports, and no coverage of the Kubernetes Operator which is the standard enterprise deployment path.

What changed:
- Restructured into two parallel sections - Apache Flink and Flink Kubernetes Operator - each with Trust Boundary, Security Boundary Reference, and Deployment Requirements subsections
- Added Security Boundary Reference tables for both components so security researchers and enterprise teams can determine in/out of scope findings without contacting the PMC 
- Documented that SSL/TLS and authentication are disabled by default (previously implied but never stated)
- Replaced FAQ format with direct statements 

CVE tracking:
The inline CVE table is removed entirely. The project does not have the bandwidth actually to keep a hand-maintained list accurate, and a stale list gives users false confidence. Going forward, CVEs are tracked exclusively through OSV (https://osv.dev) and NVD (https://nvd.nist.gov) as authoritative external databases. If the PMC decides to adopt GitHub Security Advisories as an owned record in the future, a link can be added - but that requires a process decision first.

### DIFF
--- a/docs/content/what-is-flink/security.md
+++ b/docs/content/what-is-flink/security.md
@@ -27,93 +27,76 @@ under the License.
 
 # Security
 
-## Security Updates
+## Apache Flink
 
-This section lists fixed vulnerabilities in Flink.
+Apache Flink is a distributed stream and batch processing framework that executes user-supplied code across a cluster of machines.
 
-<table class="table">
-	<thead>
-		<tr>
-			<th style="width: 20%">CVE ID</th>
-			<th style="width: 30%">Affected Flink versions</th>
-			<th style="width: 50%">Notes</th>
-		</tr>
-	</thead>
-	<tr>
-		<td>
-			<a href="https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-1960">CVE-2020-1960</a>
-		</td>
-		<td>
-			1.1.0 to 1.1.5, 1.2.0 to 1.2.1, 1.3.0 to 1.3.3, 1.4.0 to 1.4.2, 1.5.0 to 1.5.6, 1.6.0 to 1.6.4, 1.7.0 to 1.7.2, 1.8.0 to 1.8.3, 1.9.0 to 1.9.2, 1.10.0
-		</td>
-		<td>
-			Users are advised to upgrade to Flink 1.9.3 or 1.10.1 or later versions or remove the port parameter from the reporter configuration (see advisory for details).
-		</td>
-	</tr>
-	<tr>
-		<td>
-			<a href="https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-17518">CVE-2020-17518</a>
-		</td>
-		<td>
-			1.5.1 to 1.11.2
-		</td>
-		<td>
-			<a href="https://github.com/apache/flink/commit/a5264a6f41524afe8ceadf1d8ddc8c80f323ebc4">Fixed in commit a5264a6f41524afe8ceadf1d8ddc8c80f323ebc4</a> <br>
-			Users are advised to upgrade to Flink 1.11.3 or 1.12.0 or later versions.
-		</td>
-	</tr>
-	<tr>
-		<td>
-			<a href="https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-17519">CVE-2020-17519</a>
-		</td>
-		<td>
-			1.11.0, 1.11.1, 1.11.2
-		</td>
-		<td>
-			<a href="https://github.com/apache/flink/commit/b561010b0ee741543c3953306037f00d7a9f0801">Fixed in commit b561010b0ee741543c3953306037f00d7a9f0801</a> <br>
-			Users are advised to upgrade to Flink 1.11.3 or 1.12.0 or later versions.
-		</td>
-	</tr>
-	<tr>
-		<td>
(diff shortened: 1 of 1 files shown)


docs/content/what-is-flink/security.md
```diff
@@ -27,93 +27,73 @@ under the License.
...
+
+The table below is intended for security researchers and enterprise security teams evaluating Flink:
+
+| Scenario | Security boundary | Notes |
+|---|---|---|
+| Unauthenticated access to the REST API | **In scope** | Vulnerability -- report it |
+| Path traversal or unauthorized file access via REST API | **In scope** | Vulnerability -- report it |
+| SQL injection or auth bypass via SQL Gateway | **In scope** | Vulnerability -- report it |
+| Credential or secret exposure in cluster interfaces | **In scope** | Vulnerability -- report it |
+| RCE via an authenticated, submitted user job | **Out of scope** | By design -- authenticated users run arbitrary code |
```
mbalassi: might makes sense to disambiguate acronym and state `Remote Code Execution` instead of RCE.

docs/content/what-is-flink/security.md
```diff
@@ -27,93 +27,73 @@ under the License.
...
+| Spawning processes or opening connections from within a running job | **Out of scope** | By design |
+| Reading or writing files from within a running job | **Out of scope** | By design |
+
+### Deployment Requirements
+
+Flink clusters must not be exposed to the public internet. Access to all Flink network interfaces must be restricted to trusted principals via network-level controls (firewalls, security groups, VPN).
+
+SSL/TLS, REST API authentication, and SQL Gateway authentication are all **disabled by default**. A default Flink deployment is unauthenticated and unencrypted on its network interfaces. These must be explicitly configured before any production or shared deployment.
```
mbalassi: These must be explicitly configured before any production or shared deployment. -> These need to be explicitly enabled for production or shared deployments.

docs/content/what-is-flink/security.md
```diff
@@ -27,93 +27,73 @@ under the License.
...
+
+Flink does not manage the security of external systems it connects to, nor the lifecycle of credentials used to connect to them -- those are operator responsibilities.
```
mbalassi: nit: it can manage the lifecycle of tokens (not creds) - I believe the current wording is acceptable, just wanted to make sure it is not misleading.

mbalassi (approved):
Thanks, I like the direction and added some suggestions. An architecture diagram showing boundaries would be great.

MartijnVisser:
I have been discussing this subject offline with @rmetzger too, I'll try to share more in a bit

MartijnVisser:
@gaborgsomogyi @mbalassi I was iterating on top of your work, see https://github.com/MartijnVisser/flink-web/blob/expand-security-threat-model/docs/content/what-is-flink/security.md and let me know what you think

gaborgsomogyi:
@MartijnVisser thanks for the iteration. How do you think we should proceed? I've the feeling both of us have 10+ comments on the 2 versions.

gaborgsomogyi:
We can discuss the details but as a general saying I personally can't imagine a security professional who would accept the following: `Credential exposure via spec, status, events, or exceptions is Out of scope`. That would potentially mean a Flink provider could steal anybody credentials. Banks and such security teams would blow up instantly.

MartijnVisser:
> @MartijnVisser thanks for the iteration. How do you think we should proceed? I've the feeling both of us have 10+ comments on the 2 versions.

I've sent you an email, perhaps we can sync offline and then decide on a next logical step?

MartijnVisser (approved):
@gaborgsomogyi  and I synced offline. This is a clear improvement over the current page, so let's merge it as a first iteration. I'll follow up with a slimmer, higher-level pass and potentially a separate best-practices page for the configuration "how". Further input very welcome on the follow-up.

LGTM, thanks again!

docs/content/what-is-flink/security.md
gaborgsomogyi: Done.

docs/content/what-is-flink/security.md
gaborgsomogyi: Changed. Kept "configured" over "enabled" since SSL/TLS requires certificate setup beyond just flipping a flag, but adopted the rest of your phrasing.

docs/content/what-is-flink/security.md
gaborgsomogyi: Tightened. Now distinguishes short-lived tokens (Flink manages via delegation token framework) from long-lived credentials (operator responsibility).

gaborgsomogyi:
Thanks @MartijnVisser, your version highlighted important gaps. Applied the following:
- Added authenticating proxy guidance: REST API supports mutual TLS for client certificate auth; SQL Gateway has no built-in mechanism and needs a proxy in front
- Added input data deserialization as an in-scope attack surface where no Flink-level control exists to prevent it
- Tightened the RCE out-of-scope row to explicitly exclude SQL Gateway submissions, clarifying that SQL injection leading to code execution is in scope

Unless there are further comments I'm intended to merge as discussed.

I agree there are huge amount if things to do around here. One clear direction is to answer the `how`. Another one is CVE tracking in general.

PR #858, opened 2026-06-10
Replace OSV/NVD links with ASF Security page for CVE tracking

gaborgsomogyi:
Replace OSV/NVD links with ASF Security page for CVE tracking.
  
The Known Vulnerabilities section previously linked to OSV and NVD. This replaces both with the ASF Security Team's Flink-specific page at `security.apache.org/projects/flink/`.

Reasons:
- Having a targeted page for Flink CVE's is the way to go long term
- It is Flink-specific and maintained by the ASF security team with no PMC effort required
- The ASF security page is the authoritative source . NVD ingests from it, not the other way around


rmetzger (approved):
Thanks

PR #882, opened 2026-09-17
Kubernetes Operator 1.16.1 release

gyfora:
Update docs for the 1.16.1 patch release . 

I have updated the just released 1.16.0 release blogpost to not call out the reverted feature and simply updated the reference links instead of a new blogpost. Let me know if you think we should create a separate blogpost


spuru9 (commented):
Changes look fine.

But the doc may create some confusion to readers given the full doc mentions 1.16.0. 

> Could not find precedence as well. 1.12.1 never updated the post, nor created one, had a date change only still at 1.12.0. Flink creats new posts.

Atleast date changes would be nice for history tracking.

gyfora:
> Changes look fine.
> 
> But the doc may create some confusion to readers given the full doc mentions 1.16.0.
> 
> > Could not find precedence as well. 1.12.1 never updated the post, nor created one, had a date change only still at 1.12.0. Flink creats new posts.
> 
> Atleast date changes would be nice for history tracking.

I can update the doc more in detail , just wanted to avoid changing the title to not break links. In this case we really don't want people to install 1.16.0 but also showing broken links if the version is removed is not very good.

Dennis-Mircea (approved):

1996fanrui (approved):

spuru9 (approved):

